PR #5, opened 2019-06-17
Automatically restore older basebackup on repeated failure

rikonen:
It is possible for basebackups to be corrupt in some rare cases.
Depending on the issue it could possibly be resolved manually but
automatic restoration would fail. The binary logs in such cases should
be usable and restoring older basebackup and all binary logs from the
older backup and the new backup whose basebackup was unusable should
yield the same end result.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -184,16 +184,16 @@ class Controller(threading.Thread):
             else:
                 raise ValueError(f"Requested backup {stream_id!r} for site {site!r} not found")
 
-            state_file_name = os.path.join(self.state_dir, "restore_coordinator_state.json")
-            # If we're retrying restoration there could be an old state file, make sure to delete it
-            # so that obsolete state doesn't get reused
-            with contextlib.suppress(Exception):
-                os.remove(state_file_name)
             self.log.info("Restoring backup stream %r, target time %r", stream_id, target_time)
             self.state_manager.update_state(
                 mode=self.Mode.restore,
                 restore_options={
-                    "state_file": state_file_name,
+                    "binlog_streams": [{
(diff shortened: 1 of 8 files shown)


myhoard/controller.py
```diff
@@ -1196,6 +1216,45 @@ def _state_file_from_stream_id(self, stream_id):
...
+        )
+        current_stream_id = self.state["restore_options"]["stream_id"]
+        earlier_backup = None
+        for backup in backups:
+            if backup["stream_id"] == current_stream_id:
+                break
+            else:
+                earlier_backup = backup
+        else:
+            raise Exception(f"Stream {current_stream_id} being restored not found in completed backups: {backups}")
```
ojarva: It could be beneficial to raise something specific here so that if there is external python code calling myhoard directly, a specific exception could be expected for this case, instead of falling back to matching error message string.

myhoard/restore_coordinator.py
```diff
@@ -40,6 +40,8 @@ class RestoreCoordinator(threading.Thread):
     ITERATION_SLEEP_SHORT = 0.2
     ITERATION_SLEEP_LONG = 10
 
+    MAX_BASEBACKUP_ERRORS = 4
```
ojarva: This could use a short comment on telling what kind of errors (do we ignore up to MAX_BASEBACKUP_ERRORS while retrying, or is this maximum number of retries?).

rikonen: Added comment.

myhoard/controller.py
ojarva: Ah true; I was expecting this to propagate further up.

PR #15, opened 2019-08-21
tests: Always commit and flush all changes when stopping DataGenerator

rikonen:
Potentially fixes transiently failing test_restore_coordinator test in
which DataGenerator sometimes reports there should be more rows in the
target database than there actually are, which could be explained by
some rows not being flushed so they didn't get synchronized to the
target service.


rikonen:
Doesn't seem to help...

PR #20, opened 2019-09-10
controller: Remove some unnecessary logging.

juhovh-aiven:
The binlog purge and delete related logging seems to produce a
significant percentage of the total amount of logs, but it is not very
useful information.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -308,9 +308,7 @@ class Controller(threading.Thread):
                 can_purge = all(
                     stream.is_binlog_safe_to_delete(binlog, exclude_uuid=exclude_uuid) for stream in backup_streams
                 )
-                if can_purge:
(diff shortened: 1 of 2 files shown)


myhoard/controller.py
```diff
@@ -308,10 +308,6 @@ def collect_binlogs_to_purge(
                 can_purge = all(
                     stream.is_binlog_safe_to_delete(binlog, exclude_uuid=exclude_uuid) for stream in backup_streams
                 )
-                if can_purge:
-                    log.info("Binlog %s is reported safe to delete by all backup streams", binlog["local_index"])
-                else:
-                    log.info("Binlog %s reported not safe to delete by some backup streams", binlog["local_index"])
```
rikonen: I'd leave the "not safe to delete" logging, that can be informative and shouldn't occur so often.

PR #22, opened 2019-09-21
backup stream: Explicitly copy binlogs input list

rikonen:
The binlog list refers to list of all local binary logs from the
BinlogScanner class. That list must never be modified by anything else
than BinlogScanner itself to avoid unexpected side effects yet
`_upload_binlogs` used to pop items from the list. This caused
controller to fail to delete any binary logs that were not uploaded
to new backup stream until the app was restarted and restored back
to sane state.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -149,7 +149,7 @@ class BackupStream(threading.Thread):
             "mode": mode,
             "next_index": 0,
             "normalized_backup_time": normalized_backup_time,
-            "pending_binlogs": binlogs or [],
+            "pending_binlogs": list(binlogs) if binlogs else [],
             "prepare_details": {},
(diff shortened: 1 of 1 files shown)


myhoard/backup_stream.py
```diff
@@ -149,7 +149,7 @@ def __init__(
             "mode": mode,
             "next_index": 0,
             "normalized_backup_time": normalized_backup_time,
-            "pending_binlogs": binlogs or [],
+            "pending_binlogs": list(binlogs) if binlogs else [],
```
Ormod: What are the binlog objects like, i.e. do we need deepcopy for this?

rikonen: Yes and no. The binlog objects themselves should not be modified so in that sense deepcopy would be required but this is already handled in `_upload_binlog`, which takes a copy of the `dict` itself before adding some fields.

PR #25, opened 2019-10-04
Use longer than normal timeout for flushing binary logs

rikonen:
The default timeout is too low for environments under heavy load.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -18,9 +18,9 @@ from .append_only_state_manager import AppendOnlyStateManager
 from .basebackup_operation import BasebackupOperation
 from .state_manager import StateManager
 from .util import (
-    add_gtid_ranges_to_executed_set, are_gtids_in_executed_set, first_contains_gtids_not_in_second, make_fs_metadata,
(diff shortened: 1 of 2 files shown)


myhoard/backup_stream.py
```diff
@@ -778,7 +778,11 @@ def _take_basebackup(self):
             # This is to ensure the binlog is immediately backed up and available for
             # patching gtid_executed even if restoration is performed to the time when
             # basebackup was created.
-            with mysql_cursor(**self.mysql_client_params) as cursor:
+            # FLUSH BINARY LOGS might take a long time if the server is under heavy load,
+            # use longer than normal timeout here.
+            connect_params = dict(self.mysql_client_params)
+            connect_params["timeout"] = DEFAULT_MYSQL_TIMEOUT * 5
```
mwfrojdman: What's the rationale for using the constant `DEFAULT_MYSQL_TIMEOUT` as a base for multiplying instead of the timeout passed in `connect_params` (which presumably can be different from `DEFAULT_MYSQL_TIMEOUT`)?

rikonen: The client params don't have any timeout set so this used to use `DEFAULT_MYSQL_TIMEOUT` and after this change it just uses 5 times higher timeout.

PR #35, opened 2020-03-01
Handle invalid basebackup binlog position

rikonen:
The binlog position reported by XtraBackup does not always match the
actual start position of the transaction following the last GTID
included in the backup but it may be the one following that. This looks
to be locking issue on MySQL side that causes inconsistent state to be
returned.

To work around the issue add manual matching of the reported last GTID
to actual binlog position while restoring backup. It would be better to
fix this in the actual backup creation phase but that would require much
more complex logic and doing this in restoration allows coping with old
backups that are already broken.

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -21,7 +21,8 @@ from .binlog_downloader import download_binlog
 from .state_manager import StateManager
 from .util import (
     add_gtid_ranges_to_executed_set, build_gtid_ranges, change_master_to, make_gtid_range_string, mysql_cursor,
-    parse_fs_metadata, read_gtids_from_log, relay_log_name, rsa_decrypt_bytes, sort_and_filter_binlogs, track_rate
+    parse_fs_metadata, parse_gtid_range_string, read_gtids_from_log, relay_log_name, rsa_decrypt_bytes,
+    sort_and_filter_binlogs, track_rate
 )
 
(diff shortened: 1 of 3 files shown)


rikonen:
Still pending additional verification but creating pull request for review/merge in case that passes.

orange-kao:
Test passed (104 rounds). Details documented at
https://app.clubhouse.io/aiven/story/12492/mysql-insert-into-row-lost-while-doing-backup-and-upgrade-to-mysql-8-0-19

myhoard/restore_coordinator.py
```diff
@@ -952,12 +958,39 @@ def _parse_gtid_executed_ranges(self, binlog):
...
+        # file_position attributes in binlog_info don't seem to necessarily match the position that actually
+        # contains the transaction following the basebackup. This may be # related to locking in MySQL code,
+        # which could result in transaction getting written to binary log but GTID info not having been updated
+        # when table_log_status locks both binlog and gtid status, getting mismatching binlog position and gtid
+        # info.
+        last_gnos = {}
+        if self.state["basebackup_info"]["gtid"]:
+            # "gtid" contains last value for each past server so need to parse it accordingly
+            for uuid_str, ranges in parse_gtid_range_string(self.state["basebackup_info"]["gtid"]).items():
+                for rng in ranges:
```
melor: `ranges` is a sorted list, so just grab `last_gnos[uuid_str] = ranges[-1][1]` ?

PR #38, opened 2020-05-05
controller: Fix race condition in force promotion

rikonen:
If force promotion was requested while restoring binlogs and new
non-forced promotion was requested just when restore coordination
marked restoration as having finished, the force flag was not set
and promotion wasn't marked as having completed immediately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ unittest: $(generated)
 
 .PHONY: pylint
 pylint: $(generated)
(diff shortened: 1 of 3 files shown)


rikonen:
Tests are failing because of incompatibility between MySQL 8.0.20 and currently available XtraBackup. The Oracle apt repo does not seem to have older version of MySQL available so there's no obvious immediate fix for this. Since XtraBackup is likely to be updated to support this version fairly soon it may not be worthwhile to implement some workaround for getting compatible version of MySQL installed in the meantime.

PR #39, opened 2020-09-16
myhoard: On basebackup command failure raise XtraBackupError

willrouesnel-aivenio:
Identity failures of the backup process command as a specific exception type rather then just Exception, then filter them out of the normal unexpected_exception dispatch. This is because they can't raise any possible error which can be dealt with by looking at Sentry, and don't include any useful error data.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -17,6 +17,7 @@ from pghoard.rohmu.object_storage.s3 import S3Transfer
 
 from .append_only_state_manager import AppendOnlyStateManager
 from .basebackup_operation import BasebackupOperation
+from .errors import XtraBackupError
 from .state_manager import StateManager
 from .util import (
     add_gtid_ranges_to_executed_set, are_gtids_in_executed_set, DEFAULT_MYSQL_TIMEOUT, ERR_TIMEOUT,
@@ -856,10 +857,14 @@ class BackupStream(threading.Thread):
                 self.stats.gauge_float("myhoard.basebackup.compression_ratio", uncompressed_size / compressed_size)
         except Exception as ex:  # pylint: disable=broad-except
             self.log.exception("Failed to take basebackup")
-            self.stats.unexpected_exception(ex=ex, where="BackupStream._take_basebackup")
             self.stats.increase("myhoard.basebackup.errors")
             self.state_manager.increment_counter(name="basebackup_errors")
             self.last_basebackup_attempt = time.monotonic()
+            # If this is not a remote failure error (i.e. any type of error arising from programming
+            # issues in Python then we want to log this as an unexpected exception so it'll dispatch
+            # to metrics and Sentry etc.
+            if not isinstance(ex, XtraBackupError):
+                self.stats.unexpected_exception(ex=ex, where="BackupStream._take_basebackup")
         finally:
             self.basebackup_operation = None
 
--- a/myhoard/basebackup_operation.py
+++ b/myhoard/basebackup_operation.py
@@ -12,6 +12,8 @@ from contextlib import suppress
 
 from pghoard.common import increase_pipe_capacity, set_stream_nonblocking
 
+from myhoard.errors import XtraBackupError
+
 
 class BasebackupOperation:
     """Creates a new basebackup. Provides callback for getting progress info, extracts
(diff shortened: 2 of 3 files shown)


rikonen:
Static checks fail, otherwise lgtm

rikonen:
Apparently pylint version has changed and some unrelated changes are required to make that pass

willrouesnel-aivenio:
> Apparently pylint version has changed and some unrelated changes are required to make that pass

Noticed that. I think this needs to brought inline with the config we have pghoard.

willrouesnel-aivenio:
Hm, reconfiguring the travis build and it's revealed that our tests are quite racy...

willrouesnel-aivenio:
Fixed the build! The important commit is the last one: at some point the pymysql exception returned when `CHANGE MASTER TO` fails has shifted and so we weren't jumping into our failover code with these newer versions, which led to the failures when running the 3 node test. Catching the exception correctly fixes the problem.

.travis.yml
```diff
@@ -1,30 +1,111 @@
...
+  - pip install -e .
+
+inc_unittest: &inc_unittest
+  - make unittest
+
+jobs:
+  allow_failures:
+    - stage: lint
+      python: 3.6
+      env: [MYSQL_VERSION=mysql-5.7]
```
rikonen, line 39: Does setting this env value make Travis do something special or why are we setting this? It looks kind of confusing given that MyHoard only supports MySQL 8.0+

.travis.yml
```diff
@@ -1,30 +1,111 @@
...
+      before_install: *inc_before_install
+      install: *inc_install
+      script: *inc_unittest
 
+    - env: [MYSQL_VERSION=mysql-5.7]
+      python: '3.8'
+      stage: unittest
+      before_install: *inc_before_install
+      install: *inc_install
+      script: *inc_unittest
```
rikonen, line 111: Haven't been configuring Travis a lot so maybe I'm missing something but does defining exactly the same entry twice for each Python version have some effect or do these have some difference that I'm not seeing?

Makefile
```diff
@@ -5,21 +5,46 @@ generated = myhoard/version.py
 all: $(generated)
 
 PYTHON ?= python3
-PYTHON_SOURCE_DIRS = myhoard/
-PYTHON_TEST_DIRS = test/
-ALL_PYTHON_DIRS = $(PYTHON_SOURCE_DIRS) $(PYTHON_TEST_DIRS)
```
rikonen, line 10: fmt and pylint is no longer executed on test directories. At least formatting should be consistent also in tests.

Makefile
```diff
@@ -47,29 +72,3 @@ copyright:
...
-
-.PHONY: isort
-isort:
-	time isort --recursive $(ALL_PYTHON_DIRS)
-
-.PHONY: reformat
-reformat: isort yapf
-
-.PHONY: test
-test: flake8 pylint copyright unittest
```
rikonen, line 63: Copyright target is no longer executed in tests at all but would seem relevant.

.travis.yml
willrouesnel-aivenio: That is a mistake. So originally we had MYSQL 5.6, 5.7 and 8.0. Let me reinstate some of the other versions and we'll see if the full matrix works (and it'll fix up this problem).

.travis.yml
willrouesnel-aivenio: The included sections install based on this value (see `inc_install`)

Makefile
willrouesnel-aivenio: That's an oversight, I'll reinstate it.

Makefile
willrouesnel-aivenio: `PYTHON_SOURCE_DIRS` has been consolidated so it includes `test`. `fmt` and `pylint` are called against both.

.travis.yml
rikonen: Ah, right. I presume something there isn't working quite as expected and it always ends up installing MySQL 8.0 anyway. MyHoard is absolutely incompatible with older MySQL versions and will not work with those.

Makefile
rikonen: Oh yeah could've actually read the value, just assumed it didn't based on the name...

.travis.yml
willrouesnel-aivenio: Fixed.

willrouesnel-aivenio:
Everything here is rebased on the other fix, so this should be good to go now.

rikonen (approved):

PR #40, opened 2020-09-25
myhoard: Fix occasional failover problems

willrouesnel-aivenio:
At some point the pymysql error returned for ER_MASTER_INFO changed to be
an OperationalError rather then an InternalError, which means that we
weren't handling this problem properly depending on version.

Add an extra exception catch to look for both types of problem, and react
appropriately. This would only have affected newer deployments.


.travis.yml
```diff
@@ -1,30 +1,120 @@
...
+      script: *inc_unittest
+
+    - env: [MYSQL_VERSION=mysql-8.0]
+      python: '3.6'
+      stage: unittest
+      before_install: *inc_before_install
+      install: *inc_install
+      script: *inc_unittest
+
+    - env: [MYSQL_VERSION=mysql-5.7]
```
rikonen, line 94: I would double check that this actually runs the tests with MySQL 5.7. It would be surprising if it did and those worked as older MySQL versions have never been tested and afair there were several required features that were not supposed to work with the older versions.

Makefile
```diff
@@ -5,21 +5,50 @@ generated = myhoard/version.py
...
+	$(PYTHON) -m pylint --rcfile .pylintrc $(PYTHON_SOURCE_DIRS)
+
+.PHONY: fmt
+fmt: version
+	unify --quote '"' --recursive --in-place $(PYTHON_SOURCE_DIRS)
+	isort --recursive $(PYTHON_SOURCE_DIRS)
+	yapf --parallel --recursive --in-place $(PYTHON_SOURCE_DIRS)
+
+.PHONY: copyright
+copyright:
```
rikonen, line 34: copyright and coverage targets are specified twice

.travis.yml
rikonen: Though looking at it most of the version 8 specific requirements might have been outside of MyHoard but I'd still be quite surprised if 5.7 just worked out of the box.

rikonen: But also I don't know if there's any point in adding this even if it did work. We're not committed to supporting MySQL 5.7 and the very first moment any new functionality wouldn't work on 5.7 we'd just drop it and if anyone reported an issue for 5.7 we wouldn't fix it. So I think it should be removed regardless.

willrouesnel-aivenio: Not sure how I convinced myself it did, but set-selections is needed to prevent a prompt. Removed all the references to 5.7 in favor of 8.

rikonen (approved):

PR #46, opened 2021-02-19
Replace Travis build process with GitHub Actions

lornajane:
Travis has gone away, so this PR will (when finished) move the project over to using GitHub Actions.

When merged, we should switch the branch protection rules to require the new checks before merging.


lornajane:
This takes a different approach, fetching and installing specific packages from the mysql repositories manually, to enable us to specify the version to use. The version is therefore specified in this build file and will need to be bumped. I'm not sure if this is the best approach but it does mean that the tests are able to work, so I'm opening it for discussion.

rikonen (approved):
Nice! Thanks for doing this.

PR #47, opened 2021-03-03
Cleanup slave relay log info from the backup

alexole:
Old replication information from the backup might bring problems. When backup is taken from the slave, which has active replication, and restored to another standby/replica, this relay info might conflict and will end up in breaking restoration process.
    
In some cases, e.g. when rotate event appears in the relay log (coming from master binlog) and mysql has some information about non-existing replication in these tables, it tries to read previous relays logs in order to find last rotate event, but those relay logs do not exist anymore.
   
Unfortunately it's not possible to exclude these tables from the backup using xtrabackup at the moment, would have been a better option.

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -408,6 +408,9 @@ class RestoreCoordinator(threading.Thread):
                 self.log.info("No GTID ranges found in last file with given target timestamp, finalizing restore")
                 all_gtids_applied = True
 
+        if initial_round and not mysql_started:
+            self._purge_old_slave_data()
+
         self._ensure_mysql_server_is_started(**mysql_params)
 
         if not all_gtids_applied:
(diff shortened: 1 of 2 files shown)


myhoard/restore_coordinator.py
```diff
@@ -374,6 +374,7 @@ def apply_binlogs(self):
         initial_round = binlogs[0]["adjusted_remote_index"] == 1
         final_round = binlogs[-1]["adjusted_remote_index"] == self.pending_binlogs[-1]["adjusted_remote_index"]
         if initial_round and not mysql_started:
+            self._purge_old_slave_data()
```
rikonen, line 377: Is it relevant that this is done here? It would seem safer to do all the relay log manipulation while MySQL is stopped and do this just before we used to start MySQL at line 411

alexole: you are right, will change that.

myhoard/restore_coordinator.py
```diff
@@ -408,6 +408,7 @@ def apply_binlogs(self):
                 self.log.info("No GTID ranges found in last file with given target timestamp, finalizing restore")
                 all_gtids_applied = True
 
+        self._purge_old_slave_data()
```
rikonen, line 411: This should only be done the very first time this function is called.

alexole: Oops.

rikonen (dismissed):

rikonen (approved):

PR #50, opened 2021-05-03
makefile: removed duplicated targets from Makefile

NARSimoes:
Removing `copyright` and `coverage` because both targets are specified twice in `Makefile`.


rikonen:
Since you're touching this component could you check out the lint failure?

kmichel-aiven (approved):

kmichel-aiven:
Lint fixes available here : Lint fixes are here https://github.com/aiven/myhoard/pull/53

PR #52, opened 2021-06-02
remove stored state for old basebackups [BF-364]

kmichel-aiven:
Remove old .json and .remote_binlogs files related to basebackups that don't exist anymore.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1171,6 +1171,10 @@ class Controller(threading.Thread):
         backups = self.get_backup_list(
             self.backup_sites, seen_basebackup_infos=self.seen_basebackup_infos, site_transfers=self.site_transfers
         )
+        new_backups_ids = {backup["stream_id"] for backup in backups}
+        for backup in self.state["backups"]:
+            if backup["stream_id"] not in new_backups_ids:
+                self._build_backup_stream(backup).delete_state()
         self.state_manager.update_state(backups=backups, backups_fetched_at=time.time())
         return backups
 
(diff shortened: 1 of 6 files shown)


myhoard/controller.py
```diff
@@ -1171,6 +1171,10 @@ def _refresh_backups_list(self):
         backups = self.get_backup_list(
             self.backup_sites, seen_basebackup_infos=self.seen_basebackup_infos, site_transfers=self.site_transfers
         )
+        new_backups_ids = {backup["stream_id"] for backup in backups}
+        for backup in self.state["backups"]:
+            if backup["stream_id"] not in new_backups_ids:
+                self._build_backup_stream(backup).delete_state()
```
dmitry-potepalov, line 1177: Is it necessary to delete the state here? As far as I understood, the `remove()` function for old backup streams gets called on every transition to active state (inside the `_purge_old_backups` function). Does it make sense to do this cleanup in `_refresh_backups_list_and_streams` function? It already singles out new backup streams and removed backup streams.

dmitry-potepalov (changes requested):
linter complained + a small issue about cleanup for old backups

myhoard/controller.py
kmichel-aiven: A backup and a backup stream are two different things, removing the stream "closes" the backup but the backup still exists and will keep existing until it is removed by the purge process in the myhoard master (with a policy like "at most X backups or backup is at most Y days old"). 

When the backup is actually removed, then the generated list of backups won't contain references to it, and at that point only we can remove the local metadata for that removed backup.

Handling removal of metadata exactly when the list is refreshed works best because the master node knows when a backup is removed, since it initiated its removal, but the standby node merely observe the disappearance of backups and can only synchronize its metadata after observing the change.

(re the lint issue, they were pre-existing and harmless but there is a PR to fix them here: https://github.com/aiven/myhoard/pull/53 )

dmitry-potepalov: okay, thanks for the explanation)

dmitry-potepalov (approved):

PR #53, opened 2021-06-03
silence or comply with new warning from pylint 2.8.0

kmichel-aiven:
pylint introduced a consider-using-with warning to detect resource creation
not handled with a context manager.

This adds a context manager where it makes sense and silences the warning
when the code is taking responsibility of the resource by other means.


jankatins (approved):
LGTM

PR #60, opened 2021-09-17
restore: Restart sql thread if not running [BF-617]

alexanderfomichev:
In case mysql was OOM-ed/killed during the restoration process, SQL thread might not be running after restart.
To recover from that limbo state myhoard should try to restart that thread.

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -54,6 +54,13 @@ class RestoreCoordinator(threading.Thread):
     # if available (plus binlogs from the backup we were trying to restore) when this happens.
     MAX_BASEBACKUP_ERRORS = 4
 
+    # In rare cases when SQL thread needs to be restarted (mysql OOM-ed during restoration),
+    # attempts to do that should be performed with an exponential delay.
+    # The following constand defines the initial delay for the second attempt and then will be
+    # increased with a factor of 1.5 up to SQL_TREAD_RESTART_MAX_INTERVAL.
+    SQL_THREAD_RESTART_INIT_INTERVAL = 600  # 10 minutes
+    SQL_THREAD_RESTART_MAX_INTERVAL = 7200  # 2 hours
+
     @enum.unique
     class Phase(str, enum.Enum):
         getting_backup_info = "getting_backup_info"
@@ -180,6 +187,8 @@ class RestoreCoordinator(threading.Thread):
         self.target_time_approximate_ok = target_time_approximate_ok
         self.temp_dir = temp_dir
         self.worker_processes = []
+        self.sql_thread_restart_count = 0
+        self.sql_thread_restart_time = None
 
     def add_new_binlog_streams(self, new_binlog_streams):
         if not self.can_add_binlog_streams():
@@ -1128,6 +1137,27 @@ class RestoreCoordinator(threading.Thread):
                 last_renamed_index = remote_index - 1
         self.update_state(last_flushed_index=last_renamed_index, last_renamed_index=last_renamed_index)
 
+    def _start_sql_thread_if_not_running(self, slave_status):
+        # Sometimes mysqld can be OOM-killed during the backup restoration.
+        # In that case we should check if SQL thread is running and try restarting it if it is not.
+        if slave_status["Slave_SQL_Running"] == "Yes":
+            return
+
+        # Try to start SQL thread with an incremental delay
+        if self.sql_thread_restart_count > 0:
+            factor = 1.5 ** self.sql_thread_restart_count
+            iteration_delay = min(self.SQL_THREAD_RESTART_INIT_INTERVAL * factor, self.SQL_THREAD_RESTART_MAX_INTERVAL)
+            if self.sql_thread_restart_time is not None:
+                time_passed = time.monotonic() - self.sql_thread_restart_time
+                if time_passed < iteration_delay:
+                    return
+
+        self.sql_thread_restart_time = time.monotonic()
+        self.log.warning("SQL thread is not running. Running 'START SLAVE SQL_THREAD'")
+        self.sql_thread_restart_count += 1
+        with self._mysql_cursor() as cursor:
+            cursor.execute("START SLAVE SQL_THREAD")
+
     def _check_sql_slave_status(self):
         expected_range = self.state["current_executed_gtid_target"]
         expected_index = self.state["current_relay_log_target"]
@@ -1159,7 +1189,9 @@ class RestoreCoordinator(threading.Thread):
                         )
                         return True, expected_index
                 else:
+                    self._start_sql_thread_if_not_running(slave_status)
                     return False, current_index
+
             # The batch we're applying might not have contained any GTIDs
             if not expected_range:
                 self.log.info(
@@ -1185,6 +1217,9 @@ class RestoreCoordinator(threading.Thread):
                     # did not say so.
                     if expected_index is not None and current_index < expected_index:
                         current_index = expected_index
+                else:
+                    self._start_sql_thread_if_not_running(slave_status)
+
             if found:
                 cursor.execute("STOP SLAVE")
                 # Current file could've been updated since we checked it the first time before slave was stopped.


myhoard/restore_coordinator.py
```diff
@@ -1135,6 +1135,22 @@ def _check_sql_slave_status(self):
...
             cursor.execute("SHOW SLAVE STATUS")
             slave_status = cursor.fetchone()
+
+            # Sometimes mysqld can be OOM-killed during backup restoration.
+            # In that case IO and/or SQL thread might not be runnig, so they must be restarted.
+            io_thread_running = slave_status["Slave_IO_Running"]
+            sql_thread_running = slave_status["Slave_SQL_Running"]
+            refetch_slave_status = False
+            if io_thread_running != "Yes":
+                cursor.execute("START SLAVE IO_THREAD")
```
rikonen, line 1145: IO thread is typically not expected to be running and must not be automatically started.

myhoard/restore_coordinator.py
```diff
@@ -1135,6 +1135,22 @@ def _check_sql_slave_status(self):
...
+                refetch_slave_status = True
+            if sql_thread_running != "Yes":
+                cursor.execute("START SLAVE SQL_THREAD")
```
rikonen, line 1148: I'd add some backoff logic or max attempts for this. There are cases where starting the thread helps and cases where it doesn't and in the latter it will just make the situation more difficult to assess when the thread is being restarted like crazy. This condition is expected to be very rare and likely a single restart will fix it or then not so perhaps exponential backoff with initial interval of 10 minutes and exponent of 1.5 and max value of 120 minutes would be somewhat reasonable.

rikonen (dismissed):

myhoard/restore_coordinator.py
alexanderfomichev: Thanks for the explanation, will update with the delay/retry logic.

alexanderfomichev:
The change is breaking PITR test. May need some additional information to dig the reason out.

myhoard/restore_coordinator.py
```diff
@@ -1128,13 +1137,38 @@ def _rename_prefetched_binlogs_back(self, binlogs):
...
 
         with self._mysql_cursor() as cursor:
             cursor.execute("SHOW SLAVE STATUS")
             slave_status = cursor.fetchone()
+
+            # Sometimes mysqld can be OOM-killed during the backup restoration.
+            # In that case we should try restarting it.
+            if self._sql_thread_not_running(slave_status):
+                self._start_sql_thread(cursor)
+                cursor.execute("SHOW SLAVE STATUS")
```
rikonen, line 1167: I'd make `_start_sql_thread` return boolean indicating whether it started the thread or not to avoid making this call when the information is not expected to be any different.

myhoard/restore_coordinator.py
```diff
@@ -1128,13 +1137,38 @@ def _rename_prefetched_binlogs_back(self, binlogs):
...
+    def _start_sql_thread(self, cursor):
+        # Try to start SQL thread with an incremental delay
+        if self.sql_thread_restart_count > 0:
+            factor = 1 if self.sql_thread_restart_count == 1 else (self.sql_thread_restart_count - 1) * 1.5
+            iteration_delay = min(self.SQL_THREAD_RESTART_INIT_INTERVAL * factor, self.SQL_THREAD_RESTART_MAX_INTERVAL)
+            if self.sql_thread_restart_time is not None:
+                time_passed = time.monotonic() - self.sql_thread_restart_time
+                if time_passed < iteration_delay:
+                    return
+        self.sql_thread_restart_time = time.monotonic()
```
rikonen, line 1152: This should log that it does the restart

alexanderfomichev: Added some log.warning about that.

myhoard/restore_coordinator.py
alexanderfomichev: Good point! done.

rikonen (dismissed):
Didn't read very carefully last time, some more comments after reading properly

myhoard/restore_coordinator.py
```diff
@@ -1128,13 +1137,40 @@ def _rename_prefetched_binlogs_back(self, binlogs):
...
+        # Try to start SQL thread with an incremental delay
+        if self.sql_thread_restart_count > 0:
+            factor = 1 if self.sql_thread_restart_count == 1 else (self.sql_thread_restart_count - 1) * 1.5
+            iteration_delay = min(self.SQL_THREAD_RESTART_INIT_INTERVAL * factor, self.SQL_THREAD_RESTART_MAX_INTERVAL)
+            if self.sql_thread_restart_time is not None:
+                time_passed = time.monotonic() - self.sql_thread_restart_time
+                if time_passed < iteration_delay:
+                    return False
+        self.sql_thread_restart_time = time.monotonic()
+        self.log.warning("SQL thread is not running! attempting to start: START SLAVE SQL_THREAD")
```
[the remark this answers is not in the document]
alexanderfomichev: Changed to "SQL thread is not running. Running 'START SLAVE SQL_THREAD'"

myhoard/restore_coordinator.py
```diff
@@ -1128,13 +1137,40 @@ def _rename_prefetched_binlogs_back(self, binlogs):
                 last_renamed_index = remote_index - 1
         self.update_state(last_flushed_index=last_renamed_index, last_renamed_index=last_renamed_index)
 
+    def _sql_thread_not_running(self, slave_status):
+        return slave_status["Slave_SQL_Running"] != "Yes"
+
+    def _start_sql_thread(self, cursor):
+        # Try to start SQL thread with an incremental delay
+        if self.sql_thread_restart_count > 0:
+            factor = 1 if self.sql_thread_restart_count == 1 else (self.sql_thread_restart_count - 1) * 1.5
```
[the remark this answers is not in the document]
alexanderfomichev: Great nit pick, thanks!

myhoard/restore_coordinator.py
```diff
@@ -1128,13 +1137,40 @@ def _rename_prefetched_binlogs_back(self, binlogs):
...
             slave_status = cursor.fetchone()
+
+            # Sometimes mysqld can be OOM-killed during the backup restoration.
+            # In that case we should try restarting it.
+            if self._sql_thread_not_running(slave_status):
+                if self._start_sql_thread(cursor):
+                    cursor.execute("SHOW SLAVE STATUS")
+                    slave_status = cursor.fetchone()
+            else:
+                self.sql_thread_restart_count = 0
```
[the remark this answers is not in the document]
alexanderfomichev: Ok, removed. Makes sense as this code works only during syncing stage.

myhoard/restore_coordinator.py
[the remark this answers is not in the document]
alexanderfomichev: Yeah, overlooked. Thanks!

myhoard/restore_coordinator.py
```diff
@@ -1128,13 +1137,39 @@ def _rename_prefetched_binlogs_back(self, binlogs):
...
         expected_range = self.state["current_executed_gtid_target"]
         expected_index = self.state["current_relay_log_target"]
 
         with self._mysql_cursor() as cursor:
             cursor.execute("SHOW SLAVE STATUS")
             slave_status = cursor.fetchone()
+
+            # Sometimes mysqld can be OOM-killed during the backup restoration.
+            # In that case we should try restarting it.
+            if self._sql_thread_not_running(slave_status):
```
rikonen, line 1168: I checked the test failure and based on that the issue here is likely that this start is performed before checking whether the target position has been reached or not. When restoration is done in PITR mode it only restores up until certain point in a binlog and then stops the sql thread but now this goes and starts it again and it applies more transactions that is expected. Fix should be to only start the thread before returning false later on in this function (to simplify the call maybe do `_start_sql_thread` => `_start_sql_thread_if_not_running` and since the call the fetch slave status again is unnecessary you can just add simple `self._start_sql_thread_if_not_running()` call to the two relevant places later on in this function.

alexanderfomichev: Updated as suggested, hopefully I understood the code correctly.

rikonen (approved):

PR #61, opened 2021-09-17
ci: enable mypy

alexanderfomichev:
We have typechecker enabled in CI, but mypy was not installed.
Added mypy and available type info for PyMySQL and requests.
Modified a few constanst to satisfy type requirements.


rikonen:
I guess this should be rebased to fix linter

rikonen (approved):

PR #62, opened 2021-09-18
ci: calm down linter

alexanderfomichev:
There were a lot of complaints from `make lint` during builds.
This is the attempt to cleanup some:
`C0209: Formatting a regular string which could be a f-string (consider-using-f-string)`

And suppressed for now:
`W1514: Using open without explicitly specifying an encoding (unspecified-encoding)`

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -397,7 +397,7 @@ class RestoreCoordinator(threading.Thread):
             ranges = list(build_gtid_ranges(read_gtids_from_log(file_name, read_until_time=self.target_time)))
             if ranges:
                 last_range = ranges[-1]
-                until_after_gtids = "{}:{}".format(last_range["server_uuid"], last_range["end"])
+                until_after_gtids = f"""{last_range["server_uuid"]}:{last_range["end"]}"""
                 # Don't expect any specific file because if the GTID we're including is the very last entry
                 # in the file the SQL thread might switch to next file and if it is earlier then it won't
                 # so we'd need to be watching for two file names. Because execution is always single threaded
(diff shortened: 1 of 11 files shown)


myhoard/restore_coordinator.py
```diff
@@ -397,7 +397,7 @@ def apply_binlogs(self):
             ranges = list(build_gtid_ranges(read_gtids_from_log(file_name, read_until_time=self.target_time)))
             if ranges:
                 last_range = ranges[-1]
-                until_after_gtids = "{}:{}".format(last_range["server_uuid"], last_range["end"])
+                until_after_gtids = f"""{last_range["server_uuid"]}:{last_range["end"]}"""
```
rikonen, line 400: The original version was probably easier to read and I think `'` would be better substitute than `"""`.

test/test_controller.py
```diff
@@ -348,7 +348,8 @@ def streams_available_master_streaming_binlogs(backup_count):
             if backup_count > 1:
                 # Do some flushes on master to ensure there are simultaneous binlog uploads and existing ones get reused
                 with mysql_cursor(**master.connect_options) as cursor:
-                    cursor.execute("CREATE TABLE foo_{} (id INTEGER)".format(str(time.time()).replace(".", "_")))
+                    foo_prefix = str(time.time()).replace(".", "_")
```
rikonen, line 351: I guess this should be suffix instead of prefix

test/__init__.py
```diff
@@ -205,13 +205,13 @@ def random_basic_string(length=16, *, prefix=None, digit_spacing=None):
         prefix = random.choice(string.ascii_lowercase)
     random_length = length - len(prefix)
     if digit_spacing is None:
-        chars = [random.choice(string.ascii_lowercase + string.digits) for _ in range(random_length)]
+        chars = "".join([random.choice(string.ascii_lowercase + string.digits) for _ in range(random_length)])
```
rikonen, line 208: Could make do without the extra list here (`chars = "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(random_length))`).

rikonen (approved):
Few minor comments but maybe not worth spending time on those so merging anyway

PR #64, opened 2021-11-01
Add new metric which takes binlog retention period config option into…

joshua-rule:
… account

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -801,7 +801,7 @@ class Controller(threading.Thread):
         )
         # If most recent current backup is not in the list of backups being restored then we're probably
         # restoring some old backup and don't want to automatically get latest changes
-        if not any(bs["stream_id"] == backups[-1]["stream_id"] for bs in self.restore_coordinator.binlog_streams
+        if not any(bs["stream_id"] == backups[-1]["stream_id"] for bs in self.restore_coordinator.binlog_streams):
             return
 
         old_backups = [{"site": backup["site"], "stream_id": backup["stream_id"]} for backup in backups]
@@ -844,7 +844,7 @@ class Controller(threading.Thread):
             cursor.execute("SHOW SLAVE STATUS")
             info = cursor.fetchone()
             if info is None:
-                # None happens if RESET SLAVE has been perfclusterormed or if the slave never was running, e.g.
+                # None happens if RESET SLAVE has been performed or if the slave never was running, e.g.
                 # because there were no binary logs to restore.
                 self.log.warning("SHOW SLAVE STATUS returned no results.")
                 return
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1168,8 +1168,13 @@ class Controller(threading.Thread):
                 last_binlog_purge=last_purge,
                 last_could_have_purged=last_could_have_purged,
             )
+
             self.stats.gauge_float("myhoard.binlog.time_since_any_purged", time.time() - last_purge)
             self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", time.time() - last_could_have_purged)
+            self.stats.gauge_float(
+                "myhoard.binlog.time_since_should_have_purged", 
+                time.time() - (last_could_have_purged + purge_settings["min_binlog_age_before_purge"])
+            )
 
     def _refresh_backups_list(self):
         interval = self.backup_refresh_interval_base
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1078,6 +1078,17 @@ class Controller(threading.Thread):
             owned_stream_ids = [sid for sid in self.state["owned_stream_ids"] if sid != backup["stream_id"]]
             self.state_manager.update_state(owned_stream_ids=owned_stream_ids)
 
+    def _get_oldest_binlog_time(self):
+        self._process_local_binlog_updates()
+
+        oldest_binlog = 0
+
+        for binlog in self.binlog_scanner.binlogs:
+            if oldest_binlog < binlog["processed_at"]:
+                oldest_binlog = binlog
+
+        return oldest_binlog
+
     def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
         purge_settings = self.binlog_purge_settings
         if not purge_settings["enabled"] or time.time() - self.state["binlogs_purged_at"] < purge_settings["purge_interval"]:
@@ -1171,9 +1182,11 @@ class Controller(threading.Thread):
 
             self.stats.gauge_float("myhoard.binlog.time_since_any_purged", time.time() - last_purge)
             self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", time.time() - last_could_have_purged)
+
+            oldest_binlog_time = self._get_oldest_binlog_time()
             self.stats.gauge_float(
-                "myhoard.binlog.time_since_should_have_purged", 
-                time.time() - (last_could_have_purged + purge_settings["min_binlog_age_before_purge"])
+                "myhoard.binlog.time_since_oldest_should_have_purged", 
+                time.time() - (oldest_binlog_time + purge_settings["min_binlog_age_before_purge"])
             )
 
     def _refresh_backups_list(self):
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1085,7 +1085,7 @@ class Controller(threading.Thread):
 
         for binlog in self.binlog_scanner.binlogs:
             if oldest_binlog < binlog["processed_at"]:
-                oldest_binlog = binlog
+                oldest_binlog = binlog["processed_at"]
 
         return oldest_binlog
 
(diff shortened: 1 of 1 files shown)


rikonen (changes requested):
This would seemingly fix the problem but this doesn't take into account possible unexpected state that prevents deleting binlogs. If alerting was based on checking when myhoard.binlog.time_since_should_have_purged is above certain threshold then it would never fire before `min_binlog_age_before_purge` seconds had elapsed even if there was permanent problem preventing the deletion.

I'd rather update the logic so that `collect_binlogs_to_purge` returns also whether all binlogs were too recent and if they were then update `last_could_have_purged` to current time like we do if there were no binlogs with GTIDs in them.

joshua-rule:
> This would seemingly fix the problem but this doesn't take into account possible unexpected state that prevents deleting binlogs. If alerting was based on checking when myhoard.binlog.time_since_should_have_purged is above certain threshold then it would never fire before `min_binlog_age_before_purge` seconds had elapsed even if there was permanent problem preventing the deletion.
> 
> I'd rather update the logic so that `collect_binlogs_to_purge` returns also whether all binlogs were too recent and if they were then update `last_could_have_purged` to current time like we do if there were no binlogs with GTIDs in them.

@rikonen 

That's a good point.

I've changed this so the metric reports based on the oldest binlog age.
I also changed the metric so we look at the binlogs after we do the purge since we really don't care about old binlogs if the purge succeeded in deleting them.

Let me know what you think.

myhoard/controller.py
```diff
@@ -1078,6 +1078,17 @@ def _purge_old_backups(self):
             owned_stream_ids = [sid for sid in self.state["owned_stream_ids"] if sid != backup["stream_id"]]
             self.state_manager.update_state(owned_stream_ids=owned_stream_ids)
 
+    def _get_oldest_binlog_time(self):
+        self._process_local_binlog_updates()
```
rikonen, line 1082: This call seems out of place here. Is this actually needed for something?

myhoard/controller.py
```diff
@@ -1078,6 +1078,17 @@ def _purge_old_backups(self):
...
+
+        oldest_binlog = 0
+
+        for binlog in self.binlog_scanner.binlogs:
+            if oldest_binlog < binlog["processed_at"]:
+                oldest_binlog = binlog
```
rikonen, line 1088: `binlog` => `binlog["processed_at"]`

myhoard/controller.py
```diff
@@ -1168,9 +1179,16 @@ def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
...
                 last_could_have_purged=last_could_have_purged,
             )
+
             self.stats.gauge_float("myhoard.binlog.time_since_any_purged", time.time() - last_purge)
             self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", time.time() - last_could_have_purged)
 
+            oldest_binlog_time = self._get_oldest_binlog_time()
+            self.stats.gauge_float(
+                "myhoard.binlog.time_since_oldest_should_have_purged", 
+                time.time() - (oldest_binlog_time + purge_settings["min_binlog_age_before_purge"])
```
rikonen, line 1189: `time.time()` called three times here, could assign to a variable to only have it called once.

myhoard/controller.py
```diff
@@ -1168,9 +1179,16 @@ def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
                 last_binlog_purge=last_purge,
                 last_could_have_purged=last_could_have_purged,
             )
+
             self.stats.gauge_float("myhoard.binlog.time_since_any_purged", time.time() - last_purge)
             self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", time.time() - last_could_have_purged)
```
rikonen, line 1184: Should pass extra tag like `"support_should_have_purged": "1"` here so that it is easier to switch alerting to alert on either this or the new one depending on what is available.

rikonen (changes requested):
Probably fine based on very quick look, though few issues that should be fixed

myhoard/controller.py
joshua-rule: this updates the list of the binlogs so the function returns the current oldest binlog time

myhoard/controller.py
joshua-rule: good catch

myhoard/controller.py
joshua-rule: done

myhoard/controller.py
joshua-rule: added a tag

myhoard/controller.py
rikonen: Yes, I get that but the function name doesn't sound like it would be doing something like iterating through directory contents and parsing large files and I don't see why this call is needed here. We anyway periodically check the binlogs and it doesn't seem relevant whether the binlogs we have when running this code are few seconds old or not.

joshua-rule: We could rename the function. I don't really like the binlog_scanner to be honest I don't like side effects.

We do need this call, when we have a metric or a function call asking for information about something, we want the current truth not the truth as it was at some indeterminate  time in the past.

myhoard/controller.py
```diff
@@ -1163,13 +1174,23 @@ def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
...
-            self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", time.time() - last_could_have_purged)
+
+            self.stats.gauge_float("myhoard.binlog.time_since_any_purged", current_time - last_purge)
+            self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", current_time - last_could_have_purged)
+
+            oldest_binlog_time = self._get_oldest_binlog_time()
+            self.stats.gauge_float(
+                "myhoard.binlog.time_since_oldest_should_have_purged", 
+                current_time - (oldest_binlog_time + purge_settings["min_binlog_age_before_purge"]),
+                tags = { "support_oldest_should_have_purged" : 1 }
```
rikonen, line 1192: I meant adding this tag to the existing time_since_could_have_purged metric so that it is easy to not use old alert for any metric points that do have the tag present. Also would be clearer to pass the value directly as a string since tags are always strings and either it gets converted into one later or then there will be some error.

myhoard/controller.py
rikonen: OK yeah my bad I looked so hastily that didn't realize it was just purging binlogs right before sending the metric so refreshing the state is relevant. I'd be OK with just moving this call out of this function and doing it directly from the finally statement this function is getting called from.

myhoard/controller.py
joshua-rule: Ah ok made this change

myhoard/controller.py
joshua-rule: moved the call out of the function and calling it in the finally statement

myhoard/controller.py
```diff
@@ -1163,13 +1172,28 @@ def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
...
                 last_could_have_purged=last_could_have_purged,
             )
-            self.stats.gauge_float("myhoard.binlog.time_since_any_purged", time.time() - last_purge)
-            self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", time.time() - last_could_have_purged)
+
+            self.stats.gauge_float("myhoard.binlog.time_since_any_purged", current_time - last_purge)
+            self.stats.gauge_float(
+                "myhoard.binlog.time_since_could_have_purged", 
+                current_time - last_could_have_purged,
+                tags = { "support_oldest_should_have_purged" : "1" }
```
rikonen, line 1187: We're apparently not running any style checkers for this repo but should be written as `tags={"support_oldest_should_have_purged": "1"}`

rikonen (commented):
Commits need to be squashed before merging and one style comment, otherwise looks good to me

myhoard/controller.py
```diff
@@ -1078,6 +1078,15 @@ def _purge_old_backups(self):
             owned_stream_ids = [sid for sid in self.state["owned_stream_ids"] if sid != backup["stream_id"]]
             self.state_manager.update_state(owned_stream_ids=owned_stream_ids)
 
+    def _get_oldest_binlog_time(self):
```
derrix060, line 1081: nitpicking: it could be written as `sorted(lambda b: b['processed_at'])[-1]`

rikonen: Would need something a bit more complex for handling empty binlogs without index error but this actually reminds me that the binlogs are already always in sorted order. The first in the list is the oldest. So looping through the whole list is actually just wasting CPU cycles.

joshua-rule: @derrix060 I disagree having a function makes it clear what it is doing
@rikonen  premature optimization is the root of all evil =)

derrix060: Only when what is done is a bit complex.  You wouldn't be creating a function called `sum` just to add two numbers for example, or create a function called `elapsed_time` just to do what you did in the other file: `current_time - last_could_have_purged`

joshua-rule: your solution takes way longer to read and understand opposed to a function which name says exactly what it does

rikonen: Well I don't know if it's really premature optimization or just relying on expected behavior. Other code in the same class already makes that assumption:

```
        for binlog in binlogs:
            binlog_age = time.time() - binlog["processed_at"]
            min_age = purge_settings["min_binlog_age_before_purge"]
            if binlog_age < min_age:
                log.info(
                    "Binlog %s was processed %s seconds ago and min age before purging is %s seconds, not purging",
                    binlog["local_index"], math.ceil(binlog_age), min_age
                )
                break
```

And I'm not actually worried about the lost CPU cycles but rather code being confusing because it does something that seemingly shouldn't be needed and then one is left wondering which part is making assumptions it shouldn't.

But anyway can be left as it is too if you want, doesn't matter much in the end and there's already been quite a bit of iteration for this. (I could've read this properly from get go, have just looked into this briefly between handling escalations.)

rikonen (approved):

PR #73, opened 2022-03-08
controller,backup_stream: Run less frequently in case of errors

rikonen:
If we have persistent unknown errors retrying very frequently is
unlikely to make the situation better but we may end up generating a
lot of sentry errors, log spam and API requests. Increase sleep time
in case of errors for up to 30 seconds (from default 1 second).

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -428,6 +428,7 @@ class BackupStream(threading.Thread):
 
     def run(self):
         self.log.info("Backup stream %r running", self.stream_id)
+        consecutive_errors = 0
         while self.is_running:
             try:
                 if not self.file_storage:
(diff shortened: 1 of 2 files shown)


myhoard/backup_stream.py
```diff
@@ -455,12 +456,17 @@ def run(self):
...
             except Exception as ex:  # pylint: disable=broad-except
                 self.log.exception("Unexpected exception in mode %s", self.mode)
                 self.stats.unexpected_exception(ex=ex, where="BackupStream.run")
                 self.state_manager.increment_counter(name="backup_errors")
                 self.stats.increase("myhoard.backup_stream.errors")
-                time.sleep(1)
+                sleep_time = min(1.5 ** consecutive_errors, 30.0)
+                # Don't increase the counter indefinitely or else we'll get to exponent that cannot be handled anymore
+                if consecutive_errors < 20:
+                    consecutive_errors += 1
```
kmichel-aiven, line 468: nit. I'd rather avoid a "lying variable" that does not really count the errors and move the clamp to the computation instead :
```suggestion
                # Clamp the exponent to avoid OverflowError
                sleep_time = min(1.5 ** min(20, consecutive_errors), 30.0)
                consecutive_errors += 1
```

rikonen: Changed accordingly

kmichel-aiven (approved):

PR #78, opened 2022-04-27
Use python3-rohmu library instead of pghoard

raphlopez:
The [rohmu project](https://github.com/aiven/rohmu) was recently split from its parent project, [pghoard](https://github.com/aiven/pghoard). As myhoard relies on rohmu for performing backup tasks (object storage upload/download, compression, encryption), the package dependency must be updated. 

Resolves aiven/rohmu issue [#7](https://github.com/aiven/rohmu/issues/7)

### DIFF
--- a/myhoard.spec
+++ b/myhoard.spec
@@ -7,7 +7,6 @@ BuildArch:      noarch
 License:        ASL 2.0
 Source0:        myhoard-rpm-src.tar
 BuildRequires:  percona-xtrabackup-80 >= 8.0
-BuildRequires:  pghoard >= 2.0.0-115
 BuildRequires:  python3-aiohttp
 BuildRequires:  python3-devel
 BuildRequires:  python3-flake8
@@ -18,14 +17,15 @@ BuildRequires:  python3-PyMySQL >= 0.9.2
 BuildRequires:  python3-pytest
 BuildRequires:  python3-pytest-cov
 BuildRequires:  python3-requests
+BuildRequires:  python3-rohmu
 BuildRequires:  python3-socks
 BuildRequires:  python3-yapf
 BuildRequires:  rpm-build
 Requires:       percona-xtrabackup-80 >= 8.0
-Requires:       pghoard >= 2.0.0-115
 Requires:       python3-aiohttp
 Requires:       python3-cryptography >= 0.8
 Requires:       python3-PyMySQL >= 0.9.2
+Requires:       python3-rohmu >= 1.0.2
 Requires:       systemd
 
 %undefine _missing_build_ids_terminate_build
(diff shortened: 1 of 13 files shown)


ghost (dismissed):
Test failures seem related.

rikonen:
Is there some reason why this adds the rohmu dependencies but does not remove references to pghoard?

raphlopez:
There are a couple references to pghoard outside of rohmu (`pghoard.common.set_stream_nonblocking`, `pghoard.common.increase_pipe_capacity`). 
They both seem to be helper methods and are not highly coupled with anything else in pghoard. If it makes sense to copy them over into this project and remove pghoard dependency entirely, I can definitely do that.

raphlopez:
Made [PR #48](https://github.com/aiven/rohmu/pull/48) in rohmu to duplicate pghoard utils. We'll need to wait for the merge and next rohmu release before this can be merged without any pghoard dependencies.

alexole:
PR was merged, could you update this PR please? I need it to get into aiven-core for my ticket @raphlopez

myhoard/version.py
```diff
@@ -0,0 +1 @@
+__version__ = '1.0.6-45-g2ce5877'
```
alanfranz-at-work, line 1: This mustn't be committed. I don't know why this is not in .gitignore, though. I had already noticed and thought about this.

ghost (dismissed):
LGTM beyond the version glitch.

myhoard/version.py
raphlopez: Thanks for pointing that out. Removed the file and added it to the .gitignore.

requirements.txt
```diff
@@ -3,6 +3,6 @@ botocore
 cryptography >= 0.8
 httplib2
 paramiko
-pghoard >= 2.1.0
 PyMySQL >= 0.9.2
 PySocks
+rohmu
```
alexole, line 8: It won't work with any version of `rohmu`, would be better to specify `rohmu >= 1.0.2`

myhoard.spec
```diff
@@ -18,14 +17,15 @@ BuildRequires:  python3-PyMySQL >= 0.9.2
...
+BuildRequires:  python3-rohmu
 BuildRequires:  python3-socks
 BuildRequires:  python3-yapf
 BuildRequires:  rpm-build
 Requires:       percona-xtrabackup-80 >= 8.0
-Requires:       pghoard >= 2.0.0-115
 Requires:       python3-aiohttp
 Requires:       python3-cryptography >= 0.8
 Requires:       python3-PyMySQL >= 0.9.2
+Requires:       python3-rohmu
```
alexole, line 28: Same here and above (see my comment about the version)

alexole (approved):
LGTM

PR #79, opened 2022-05-05
Prevent restores from continuing indefinitely

RommelLayco:
# About this change: What it does, why it matters

In mysql versions greater than 8.0.26, the default terminology changed. See WL#14189 and WL#14628. The Slave_SQL_Running_State message has changed to use Replica instead of Slave. This change ensures that
myhoard can handle both terms.

See
https://dev.mysql.com/worklog/task/?id=14189
https://dev.mysql.com/worklog/task/?id=14628

 https://github.com/mysql/mysql-server/commit/2e273ed403827fa3d86970b40280d4fe9d154c43

Issue #77, opened 2022-04-05
Intermittent build workflow failures


st3fan:
Quick note to mention that I got a number of intermittent test errors on the two PRs I submitted today. I tried to discover the root cause but my knowledge of MySQL and these tests is limited. I did re-run the failed jobs and eventually they all succeeded.

alanfranz-at-work:
This should be fixed with #82 . If you still encounter intermittent failures, it's at least easier to check what's happening on your own workstation and see if it's just flaky infrastructure or a real problem.

PR #82, opened 2022-05-11
 build system overhaul - use container images for testing

ghost:
build system overhaul - use container images for testing
    
While investigating an environment-dependent bug, I found it very hard
to reproduce the testing environment, since it was using github actions
with their own virtual machines content.
    
This approach should make debugging much faster on developer
environments: github actions uses a docker image for testing, and the
very same image can be built and used from the developer machine. The
github workflow only contains calls to scripts that can be executed in
the same way on a developer's machine.

This also fixes the issues that were detected in #77 and while trying to merge #79


.github/workflows/build.yaml
```diff
@@ -82,68 +111,56 @@ jobs:
...
   unittest:
+    needs: build-and-push-image
     runs-on: ubuntu-latest
+    container: 
+      image: ghcr.io/aiven/myhoard-test:${{ github.sha }}
     strategy:
+      fail-fast: false
+      # this isn't a standard matrix because some version testing doesn't make sense
+      # e.g. you cannot use mysql .28 with percona .27, and there's no clean way of
+      # skipping matrix items that aren't meaningful
```
kmichel-aiven, line 122: This should be possible: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude

kmichel-aiven, line 122: This should be possible: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude

alanfranz-at-work: Technically, it could be possible. But would it be better? Then you need to calculate what's being executed and what not. This approach just uses a "flat list" instead of a matrix, and I doubt we'd need to check all possible python versions (mysql vs percona versions seem fare more risky!)

Makefile
```diff
@@ -70,3 +70,33 @@ build-dep-fedora:
...
+
+# when the image didn't change this can be used. local dev only, don't use in CI
+# in this target we override the /src that gets used to rsync source inside the container
+.PHONY: dockertest-resync
+dockertest-resync:
+	docker run -it --rm -v "$(shell pwd):/src:ro" myhoard-test-temp /src/scripts/test-inside
+
+.PHONY: dockertest-pytest
+dockertest-pytest:
+	docker run -it --rm -v "$(shell pwd):/src:ro" myhoard-test-temp /src/scripts/pytest-inside $(PYTEST_ARGS)
```
kmichel-aiven, line 102: Missing newline at end of file.

README.md
```diff
@@ -745,6 +745,33 @@ Current phase of backup restoration. Possible options are these:
   automatically but it may fail repeatedly and analyzing logs to get more
   details regarding the failure is advisable.
 
+Running tests
+=============
```
kmichel-aiven, line 749: Can you add a section about running the tests without docker too ?

build-image/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:focal
+ENV DEBIAN_FRONTEND="noninteractive"
+RUN apt-get update && apt-get -y install sudo lsb-release wget tzdata libsnappy-dev \
+    libpq5 libpq-dev software-properties-common build-essential rsync curl
+RUN add-apt-repository -y ppa:deadsnakes/ppa
```
kmichel-aiven, line 5: Missing newline at end of file.

scripts/create-user
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
+adduser --disabled-password --gecos "" testme || true
+adduser testme adm
+adduser testme sudo
+echo "%sudo ALL=(ALL:ALL) NOPASSWD: ALL" > /etc/sudoers.d/sudonopasswd
+chmod 440 /etc/sudoers.d/sudonopasswd
+chown -R testme:testme .
```
kmichel-aiven, line 7: Missing newline at end of file.

test/test_controller.py
```diff
@@ -285,7 +285,7 @@ def test_backup_state_from_removed_site_is_removed(default_backup_site, mysql_em
     for file_name in fake_file_names:
         assert not os.path.exists(file_name)
 
-
+@pytest.mark.skip(reason="This seems very flaky. See BF-1198")
```
kmichel-aiven, line 288: I don't think we should have references to BF-* in this repo.

test/test_util.py
```diff
@@ -255,10 +257,19 @@ def test_encrypt_decrypt():
...
 
 
-def test_detect_running_process_id():
-    pytest_id = myhoard_util.detect_running_process_id("python3 -m pytest")
-    coverage_id = myhoard_util.detect_running_process_id("python3 -m coverage")
-    if pytest_id is None and coverage_id is None:
-        output = subprocess.check_output(["ps", "-x", "--cols", "1000", "-o", "pid,command"])
-        raise AssertionError(f"None of the commands included in ps output {output!r}")
-    assert myhoard_util.detect_running_process_id("certainlynosuchprocesscurrentlyrunning") is None
+class TestDetectRunningProcessId:
```
kmichel-aiven, line 260: pytest does not recommend xunit-style test classes with setup/teardown functions, using fixtures is more native to the framework and composes better. (it's not possible to ask another fixture from the setup/teardown functions for instance).

.github/workflows/build.yaml
```diff
@@ -1,7 +1,5 @@
 name: Build MyHoard
 
-permissions: read-all
-
```
kmichel-aiven, line 4: This reverts the hardening PR: https://github.com/aiven/myhoard/pull/75
I doubt that's a good thing.

kmichel-aiven (dismissed):
A few small changes.

The PR should be renamed too, it's not a reproducible image: there are apt-get and wget inside the docker scripts that result in a different image every time it's built.

.github/workflows/build.yaml
alanfranz-at-work: This change was originally done to for the junit publish, for which I had got an explicit "go" from security. But I think this is unnecessary now, so, good catch.

Also, note that #75 says:
```Set the default workflow permissions to read - because this workflow does not need to modify anything in the project.```

So, it's a correct security-by-default measure, but it was "hardening", not "fixing a security problem". If we need to relax permissions for good reasons, I see nothing preventing this.

Makefile
alanfranz-at-work: Do we have any standard about this? I'd feel silly to check and fix this manually. ```make lint``` doesn't catch this either.

README.md
alanfranz-at-work: Makes sense, will do.

test/test_controller.py
alanfranz-at-work: uh, OK, that I didn't know, will double-check. Is this only for source code? Or for PRs/commit messages as well?

test/test_util.py
alanfranz-at-work: I suppose you're referring to this:

```While these setup/teardown methods are simple and familiar to those coming from a unittest or nose background, you may also consider using pytest’s more powerful [fixture mechanism](https://docs.pytest.org/en/6.2.x/fixture.html#fixture) which leverages the concept of dependency injection, allowing for a more modular and more scalable approach for managing test state, especially for larger projects and for functional testing. You can mix both fixture mechanisms in the same file but test methods of unittest.TestCase subclasses cannot receive fixture arguments.```

I'd not say it's not recommended, just that it's an alternative approach.

"more native to the framework"  -> doesn't mean better, though. What advantage would we get from this 
"composes better" -> we don't need composition here, it's a very small test, and I think I've seen plenty of examples where fixtures are abused and reused too many times, and their dependency chain can lead to an hard-to-identify test behaviour.

My take would be "use a fixture where a fixture is useful", not "use a fixture as the default approach". The test as written is very self-contained and quite easy to understand. I see no advantage in using a fixture with a cleanup action.

alanfranz-at-work:
> The PR should be renamed too, it's not a reproducible image: there are apt-get and wget inside the docker scripts that result in a different image every time it's built.

It's reproducible between the developer's environment and github actions. But I'm always very attentive to naming issues, so, I'll try to come up with a better naming and description.

Makefile
kmichel-aiven: pylint does it for python files (C0304), but we don't have a linter for non-python.
the standard is... [posix](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) :D (and GitHub showing a red sign on the line)

Most text editors and IDEs have an option to fix that automatically instead of waiting for a linter:
 - Pycharm: Settings→Editor→General→Ensure every saved files ends with a line break (at the bottom)
 - VsCode: File→Preferences→Settings→Search for "insert final newline"
 - VIM: `:set fixendofline`

alanfranz-at-work: Yeah, I know that it's posix, but I'm not sure we should focus on this kind of details in a review. This should be catched (and fixed!) by some automatic tool. I'll change those right now, btw.

kmichel-aiven (approved):

PR #83, opened 2022-05-16
misc: make sure all files have a final newline

kmichel-aiven:

Check and fix all files that are missing a final newline, so that we don't need to check it manually.


ghost (approved):
Uhm, so it requires a custom check. 

I'm merging this, because it's undoubtedly useful and it's an improvement, but I must admit I don't use the 'fmt' target because it doesn't work for me, neither on Fedora or on Ubuntu (I hadn't time to check why, it seems **unify** doesn't support the proper flags).

PR #84, opened 2022-05-16
misc: replace unify/yapf with black 

kmichel-aiven:

Replace unify (which didn't really work) and yapf, with black.


ghost (approved):
LGTM.

I'll take care of checking if we've got a kind of "formatting template" for all non-core projects; we shouldn't need to do this separately for every (python) project.

Thanks, Kevin.

PR #85, opened 2022-05-16
detect_running_process_id fix signature and return values

ghost:
* Fix func signature
* Fix return value in miss path


alanfranz-at-work:
> Test failure seem related

totally. I tried too many things at once :-/

kmichel-aiven (approved):

PR #86, opened 2022-05-20
Add missing build dependency `systemd-rpm-macros` into rpm spec file

alexole:
RPM package build requires `systemd-rpm-macros` which is missing in the spec file. So adding it.


kmichel-aiven (approved):

alanfranz-at-work:
I'm trying to add a fedora test build so we can build the rpm in CI as well.https://aiven.atlassian.net/browse/BF-1200

PR #87, opened 2022-06-02
Report binlog_upload_delay to better track problems with binlog uploads

nicois:
This adds a `binlog_upload_delay` property of `BackupStreams` to report a value of 0 when there are no eligible binlogs to stream. If there are pending binlogs, it reports the maximum age of these files, based on their `mtime`.

This metric will assist in identifying problems with transferring pending binlogs, as the existing `binlog_upload_age` metric will reach arbitrarily high values even when no binlogs are available for upload.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -258,6 +258,19 @@ class BackupStream(threading.Thread):
             return 0
 
     @property
+    def binlog_upload_delay(self) -> int:
+        """Returns the number of seconds since the oldest pending binlog was written."""
+        now = time.time()
+        oldest_mtime = now
+        for binlog in self.state["pending_binlogs"]:
+            try:
+                oldest_mtime = min(oldest_mtime, os.stat(binlog["full_name"]).st_mtime)
+            except FileNotFoundError:
+                pass  # binlog was just purged, so ignore it.
+
+        return int(now - oldest_mtime)
+
+    @property
     def created_at(self):
         return self.state["created_at"]
 
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1410,6 +1410,11 @@ class Controller(threading.Thread):
         self.stats.gauge_float("myhoard.binlog_upload_age_min", min(binlog_upload_ages))
         self.stats.gauge_float("myhoard.binlog_upload_age_max", max(binlog_upload_ages))
 
+        # Send additional info about how long 'complete' binlog files have been waiting to be uploaded
+        binlog_upload_delays = [backup_stream.binlog_upload_delay for backup_stream in backup_streams]
+        self.stats.gauge_int("myhoard.binlog_upload_delay_min", min(binlog_upload_delays))
+        self.stats.gauge_int("myhoard.binlog_upload_delay_max", max(binlog_upload_delays))
+
     def _set_uploaded_binlog_references(self):
         references = self.state["uploaded_binlogs"]
         for reference in references:


myhoard/backup_stream.py
```diff
@@ -248,12 +248,18 @@ def add_remote_reference(self, *, local_index, remote_key):
...
+           binlog was written, ignoring the still-active binlog file -
+           but only if the binlog is not fully uploaded."""
+        if self.is_streaming_binlogs() and not self._is_up_to_date():
+            return time.time() - self.state["last_binlog_upload_time"]
+
+        # Either we're not supposed to be streaming binlogs,
+        # or all eligible binlogs have been uploaded.
+        return 0
+
+    def _is_up_to_date(self) -> bool:
```
songnon, line 261: `pending_binlog_count` can be used directly, I guess.

rikonen:
The fact that nothing has been uploaded in a long time can still be a problem in itself even if there is a legitimate reason for it in some special case. Perhaps the logic for rotating binlogs is broken and nothing is being uploaded even though something should be. I'd consider e.g. adding a tag to the metric telling whether pending binlogs exist or not or adding some completely new metric, whichever would allow having more relaxed alert created for the cases where it's maybe OK that nothing has been uploaded in a longer time without completely removing the information about when something was actually uploaded.

nicois:
> The fact that nothing has been uploaded in a long time can still be a problem in itselfe information about when something was actually uploaded.

Agreed. Upon reflection it does make sense to be a different metric. I'll amend this accordingly.

myhoard/backup_stream.py
```diff
@@ -255,6 +255,15 @@ def binlog_upload_age(self):
             # If we're not supposed to be streaming binlogs then report age as 0
             return 0
 
+    @property
+    def binlog_upload_delay(self):
+        """Returns the number of seconds since the oldest pending binlog was written."""
+        mtimes = {os.stat(binlog["full_name"]).st_mtime for binlog in self.state["pending_binlogs"]}
```
alanfranz-at-work, line 261: I think this can lead to a race condition if the binlog is removed from disk after being read from state but before being statED. I cannot check which thread is doing what right now, but his should be taken into account and the exception on stat ignored.

myhoard/controller.py
```diff
@@ -1389,6 +1389,11 @@ def _send_binlog_stats(self):
         self.stats.gauge_float("myhoard.binlog_upload_age_min", min(binlog_upload_ages))
         self.stats.gauge_float("myhoard.binlog_upload_age_max", max(binlog_upload_ages))
 
+        # Send additional info about how long 'complete' binlog files have been waiting to be uploaded
+        binlog_upload_delays = [backup_stream.binlog_upload_delay for backup_stream in backup_streams]
+        self.stats.gauge_float("myhoard.binlog_upload_delay_min", min(binlog_upload_delays))
```
alanfranz-at-work, line 1394: I'm unsure about the usefulness about the minimum upload delay. But if you think it's OK, it's fine for me to add, it won't break anything.

ghost (dismissed):

myhoard/backup_stream.py
nicois: ok, done.

myhoard/controller.py
nicois: Yes, I think it is OK. Thanks!

alanfranz-at-work, line 1394: Minor nitpicking, sorry, I hadn't seen this before.

You're using gauge_float but you seem to return an int from  binlog_upload_delay, so, the types don't match. You should choose the proper gauge_ function.

Could you please also add the return type annotation in the binlog func?

ghost (dismissed):

myhoard/backup_stream.py
```diff
@@ -257,6 +257,21 @@ def binlog_upload_age(self):
             # If we're not supposed to be streaming binlogs then report age as 0
             return 0
 
+    @property
+    def binlog_upload_delay(self) -> int:
+        """Returns the number of seconds since the oldest pending binlog was written."""
+        mtimes = set()
+        for binlog in self.state["pending_binlogs"]:
+            try:
+                mtimes.add(os.stat(binlog["full_name"]).st_mtime)
```
alexole, line 266: why not just
```
min_mtime = min(os.stat(binlog["full_name"]).st_mtime, min_mtime)
```
and use it later

nicois: This is in a `try` .. `except` block as the file may no longer be present.

alexole: I mean there is no point in keeping all mtimes in a set, when you need only one value.

nicois: Oh I see. Sure, I will make that change. Although this is a local variable, it's good practice to bound memory usage.

facetoe (approved):

PR #88, opened 2022-06-03
Emit binlog upload rate metric to statsd

nicois:
Usually, binlogs are relatively small and the time taken to upload to S3 is negligible. However, in some situations (e.g. long-running transactions), binlogs will not be finalised as quickly, leading both to larger binlog files, as well as greater time increments between successive binlogs. This in itself is not a problem, but the existing metrics do not make it easy to differentiate between long-running transactions, and actual problems with processing binlogs.

One part of a robust solution involves exposing a metric which provides assurance that the S3 upload is occurring, even if the biglog is taking a while to be transferred. With this present, an alerting system can ignore metrics relating to binlogs not being sent if the current transfer rate is above a certain threshold.

The impementation involves the creation of a long-running thread, attached to the controller. This allows for a single metric to be emitted which collates all binlog transfers across all `BackupStream` instances. The pre-existing `trace_rate` function is limited in a few ways, both regarding not handling low transfer speeds, as well as performing crude bitrate calculations based on successive data points.

### DIFF
--- a/myhoard/util.py
+++ b/myhoard/util.py
@@ -3,7 +3,9 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import padding
 from cryptography.hazmat.primitives.hashes import SHA1
-from typing import Optional, Tuple
+from logging import Logger
+from math import log10
+from typing import List, Optional, Tuple
 
 import collections
 import contextlib
@@ -16,6 +18,7 @@ import socket
 import struct
 import subprocess
 import tempfile
+import threading
 import time
 
 DEFAULT_MYSQL_TIMEOUT = 4.0
@@ -382,7 +385,7 @@ def sort_and_filter_binlogs(*, binlogs, last_index, log, promotions):
         if last_range_start is not None:
             ranges.append([last_range_start, range_start - 1, promotions[last_range_start]])
         last_range_start = range_start
-    ranges.append([last_range_start, 2 ** 31, promotions[last_range_start]])
+    ranges.append([last_range_start, 2**31, promotions[last_range_start]])
 
     binlogs.sort(key=lambda bl: (bl["remote_index"], bl["server_id"]))
     valid_binlogs = []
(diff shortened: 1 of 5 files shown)


myhoard/util.py
```diff
@@ -469,3 +470,49 @@ def track_rate(*, current, last_recorded, last_recorded_time, metric_name, min_i
...
+        self._running = True
+
+    def increment(self, value: float) -> None:
+        now = time.monotonic()
+        with self._previous_values_lock:
+            self._previous_values.append((now, value))
+
+            # delete old values until the oldest value is "just" over 'window' seconds ago
+            while len(self._previous_values) > 1 and self._previous_values[1][0] + self.window < now:
+                self._previous_values.pop(0)
```
kmichel-aiven, line 501: I'm worried about the memory usage (total amount and general overhead from list manipulation) if the network speed is large and/or packets are small and/or the window is large.

It could be safer to do an exponential averaging :
```python
time_constant_secs = 4
delta_time_secs = now - last_update
smooth_value = value - math.exp(-delta_time_secs/time_constant_secs) * (value - smooth_value)
```
A time constant of 4 takes 59sec to go from 1MB/sec to ~0, 86sec to go from 1GB/sec to ~0, 114sec to go from 1TB/sec to ~0.

But if we want a perfectly clean graph (that goes from max to zero between two ticks of the frequency, after rounding of the value), we can pick it from the update frequency and the maximum expected transfer rate:

```python
time_constant_secs = (frequency_secs - math.log(2)) / math.log(max_speed_bytes)
```
For 30sec and 1GB/sec, we get a time constant of 1.4.

kmichel-aiven (changes requested):

myhoard/util.py
nicois: Just a brief reply/thought: one possibility to ensure an upper bound on memory use is to round each timestamp by (say) window/10. If the new value has the same adjusted timestamp as the latest value, we combine them.

This would result in at most 10 (or 11) items in the list, and would still provide a fairly "snappy" response to variations in transfer rates.

kmichel-aiven (approved):

kmichel-aiven:
Small thing, remember to dismiss the old review after addressing changes so that it can reappear as review required.

PR #89, opened 2022-06-05
Gracefully handle failure to purge binlogs during backup

nicois:
If a backup is in progress, deletions of binlogs will fail. This should be considered a 'normal' situation, and should be handled more gracefully.

A sample error message emitted by the mysql client:
`pymysql.err.OperationalError: (4085, 'Could not purge binary logs since another session is executing LOCK INSTANCE FOR BACKUP. Wait for that session to release the lock.')`

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -36,6 +36,7 @@ import threading
 import time
 
 ERR_CANNOT_CONNECT = 2003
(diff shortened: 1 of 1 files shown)


ghost (approved):
Looks OK to me; if doing this again please separate the commit doing the "real thing" and the commit inverting the if/else condition and blocks.

PR #90, opened 2022-06-05
Expose a metric measuring the delay applying new binlogs.

nicois:
To enable more targeted alerting, an additional metric should be emitted which measures the time taken to apply any new available binlogs.

While there are metrics related to replication lag, these are not helpful as they do not differentiate between legitimate causes (e.g. locked tables in either the primary or replica), and do not assist with identifying whether the problem is in transmitted the binlogs from the primary to S3, or from S3 to the replica.


nicois:
I am not happy with the current approach. I need to review my understanding of the myhoard states, and what data is expected to be transferred in each mode (in particular, observer and restore).

PR #91, opened 2022-06-07
Provide a basic pre-commit configuration

nicois:
A packaged pre-commit configuration file will assist contributers in quickly verifying their code complies with black/mypy/pylint/etc.

Among other benefits, the checks are more targeted, only validating files changed in the commit.

Github CI is updated to use the same tooling.


kmichel-aiven (dismissed):
```
This does not affect the CI steps executed in github.
```
I think it should, like pghoard, it seems better to keep a single codepath for linters used locally and in CI. Divergence that creep-up over time and make a PR fail only in CI can be frustrating for contributors .

kmichel-aiven (approved):

PR #94, opened 2022-10-30
Ensure that basebackup progress is set to 100 upon success.

nicois:
Because of how the `myhoard.basebackup.estimated_progress` metric is calculated, the final value was not always 100 when a backup is successfully generated.

This does not alter how backups work, just adds a final transmission of the progress metric after a basebackup has completed.

This is important as it allows prometheus metrics to accurately reflect whether a backup was successful.

### DIFF
--- a/myhoard/basebackup_operation.py
+++ b/myhoard/basebackup_operation.py
@@ -117,7 +117,7 @@ class BasebackupOperation:
                         self._process_input_output()
 
         self.data_directory_size_end, self.data_directory_filtered_size = self._get_data_directory_size()
-        self._update_progress()
+        self._update_progress(estimated_progress=100)
 
     def _get_data_directory_size(self):
         total_filtered_size = 0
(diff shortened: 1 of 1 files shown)


aris-aiven (dismissed):
LGTM

nicois:
> This depends on #95 and will be merged in lockstep unless someone objects.

I've created this PR to have #95 as a target. If it's merged to master first, this PR will automatically be modified to merge to `master`.
I've also just pushed this commit, rebased off #95.

myhoard/basebackup_operation.py
```diff
@@ -118,6 +118,7 @@ def create_backup(self):
 
         self.data_directory_size_end, self.data_directory_filtered_size = self._get_data_directory_size()
         self._update_progress()
+        self.stats.gauge_float("myhoard.basebackup.estimated_progress", 100)
```
packi, line 121: nit: I would probably inject the `100` into `_update_progress` (and not calculate it if it was provided) to avoid duplicating the stat name

aris-aiven (approved):
LGTM

PR #95, opened 2022-10-31
Update CI and local test harness to resolve regressions

nicois:
Currently the `master branch is failing, as the distribution no longer behaves as expected.
While identifying the causes of the regressions, a number of opportunities for streamlining the CI process were seen and acted upon, to streamline the way in which we use github actions, as well as how we construct local docker-based environments for unit testing.

This PR does not change anything about myhoard itself, only how we test it.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,4 +1,4 @@
-name: Build MyHoard
+name: Unit tests
 
 permissions: read-all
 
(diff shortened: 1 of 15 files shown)


aris-aiven (dismissed):
Installation workflow is clear and cleanups are appreciated.
If no one requests changes within a day I'll merge it.

aris-aiven (approved):

alanfranz:
Hello @nicois @aris-aiven , I just discovered this PR reverted some changes to the build workflow using docker images, which instead I was going to push as the "way to go" for other projects as well.

The problem with using raw github actions is that it's hard to say what's happening. Things are running in a github-provided VM which you have no control of, and using actions that install versions that you don't know about; it's hard to reproduce the testing environment in that scenario, so you get flaky tests that you can't debug.

The docker-based setup gives environment parity between developer machines and github actions.

The reasoning for using images is cleary described in the commit that introduced them: https://github.com/aiven/myhoard/commit/5187809a367585b9f3518d127a253cdf135628ca and the change was quite recent; so, while you can disagree with it, I think asking would have been better :-)

alanfranz:
Don't worry, we'll fix that; I'm starting a discussion about this topic shortly.

 One additional remark: the build is currently using a third-party action for caching APT packages, which is not a github-provided action and I couldn't find it within verified creators https://aiven.slab.com/posts/git-hub-actions-9dkctxd8  @nicois did you get clearance from security team to use that action?

PR #96, opened 2022-11-03
Transmit metric on backup initiation

nicois:
This will assist in more easily monitoring basebackups. By sending a metric as soon as the basebackup creation process begins, it can be used to reconcile subsequent activity, ensuring that it is closely followed by basebackup progress metrics.


aris-aiven (approved):
All good

PR #97, opened 2022-11-03
Ensure package can be built and tested with Python 3.11

aris-aiven:
Python 3.11 was released recently, and myhoard aims to support it.


nicois:
Can we remove support for older python versions too? At least allowing us to use python3.8 syntax would be nice. 
Versions of python prior to 3.10 are only receiving security updates; it would be reasonable for us to do likewise.

https://devguide.python.org/versions/

myhoard/controller.py
```diff
@@ -353,7 +353,7 @@ def switch_to_observe_mode(self):
                 return
             elif self.mode in {self.Mode.active, self.Mode.promote}:
                 # Master (or almost master) cannot become a standby
-                raise ValueError(f"Switch from {self.mode} to observe mode is not allowed")
+                raise ValueError(f"Switch from {self.mode.value} to observe mode is not allowed")
```
kmichel-aiven, line 356: Why this change ?

version.py
```diff
@@ -1,50 +1,65 @@
 # Copyright (c) 2019 Aiven, Helsinki, Finland. https://aiven.io/
```
kmichel-aiven, line 1: I can guess the reason for some of the changes (imp deprecated, etc.), but I'm not sure about the reason behind others ?

kmichel-aiven (dismissed):
The changes are probably all good, but most of them are not described and mixed together in the same commit, this should be split and explained a little more.

myhoard/controller.py
aris-tritas-aiven: This is due to a change to Enum.format behaviour (see https://docs.python.org/3/whatsnew/3.11.html):

> Changed [Enum.__format__()](https://docs.python.org/3/library/enum.html#enum.Enum.__format__) (the default for [format()](https://docs.python.org/3/library/functions.html#format), [str.format()](https://docs.python.org/3/library/stdtypes.html#str.format) and [f-string](https://docs.python.org/3/glossary.html#term-f-string)s) of enums with mixed-in types (e.g. [int](https://docs.python.org/3/library/functions.html#int), [str](https://docs.python.org/3/library/stdtypes.html#str)) to also include the class name in the output, not just the member’s key. This matches the existing behavior of [enum.Enum.__str__()](https://docs.python.org/3/library/enum.html#enum.Enum.__str__), returning e.g. 'AnEnum.MEMBER' for an enum AnEnum(str, Enum) instead of just 'MEMBER'.

aris-tritas-aiven: I chose to keep the same error message.

kmichel-aiven:
> Can we remove support for older python versions too? At least allowing us to use python3.8 syntax would be nice. Versions of python prior to 3.10 are only receiving security updates; it would be reasonable for us to do likewise.
> 
> https://devguide.python.org/versions/

Removing 3.8 and older sounds reasonable.
I wouldn't remove 3.9 since it's still the current version in the latest stable debian (bullseye) and latest RHEL (9).

version.py
aris-tritas-aiven: In the specific file you mean? The main change is going from the ``imp`` module to ``importlib``.
I fixed pylint errors and added print statements for traceability (where do we get the version string from?).
The type hints and variable renaming were not necessary.

myhoard/controller.py
kmichel-aiven: This makes sense, but this should be in the commit messages :stuck_out_tongue:

aris-tritas-aiven: Done!

kmichel-aiven (approved):

PR #99, opened 2022-11-18
Cope with "SHOW SLAVE STATUS" returning empty output

jelmer:


aris-aiven:
Do we know if the failing combination should be removed from the test matrix?

rikonen:
Commit message should clarify what is the scenario this happens in and why this fixes it. I'd imagine `SHOW SLAVE STATUS` should only return empty result in case `RESET SLAVE` has been called but if that is the case then no more data would be restored so this would just cause the code not to trip on access of `None` object but it would not be in a good state either (unless there's something that fixes this state later on in the call path, didn't really check).

jelmer:
Sorry, didn't mean for this to show up on anybody's radar yet because it's not ready - created this PR as draft to check whether it'd trip any tests.

PR #103, opened 2022-11-25
Publish to PyPI with GitHub Action

alexey-tereshenkov-aiven:
The project is not currently published to PyPI which makes it impossible to use in a standard virtual environment setup.
This PR adds a GitHub Action that publishes sdist and wheel to PyPI based on pushed annotated tag. The code is copied from the [aiven-client action](https://github.com/aiven/aiven-client/blob/main/.github/workflows/publish-pypi.yaml).

The secret `secrets.PYPI_API_TOKEN` needs to be set up. Could someone please help me with that? Thank you!

### DIFF
--- a/.github/workflows/publish-pypi.yaml
+++ b/.github/workflows/publish-pypi.yaml
@@ -23,7 +23,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.10
+        python-version: "3.10"
 
     - name: Install pypa/build
       run: >-
--- a/.github/workflows/publish-pypi.yaml
+++ b/.github/workflows/publish-pypi.yaml
@@ -1,0 +1,48 @@
+# Based on https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+
+name: Publish to PyPI
+
+permissions: read-all
+
+on:
+  push:
+    tags:
+      - '**'
+
+jobs:
+  build-n-publish:
+    name: Build and publish
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        persist-credentials: false
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.10
+
+    - name: Install pypa/build
+      run: >-
+        python -m pip install build --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+
+    - name: Verify tag is annotated
+      if: startsWith(github.ref, 'refs/tags')
+      run: >-
+        test $(git for-each-ref --format='%(objecttype)' ${GITHUB_REF}) == tag
+    - name: Publish distribution to PyPI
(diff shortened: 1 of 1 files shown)


alanfranz:
> makes it impossible to use in a standard virtual environment setup.

What is your use case? What is your setup? If you ```pip install --editable .``` it's certainly usable in venv.

I don't think we currently publish non-libraries to pypi (e.g. we publish rohmu) but I don't see a real reason why we shouldn't.

ccing @aiven/aiven-open-source-reviews for opinions and @aiven/team-security for secrets configuration info.

alexey-tereshenkov-aiven:
> > makes it impossible to use in a standard virtual environment setup.
> 
> What is your use case? What is your setup? If you `pip install --editable .` it's certainly usable in venv.

Thanks for looking at this! I don't have a concrete use case yet, but being able to bootstrap an environment from `requirements.txt` may be helpful (for me and others) instead of cloning the GitHub repository. That's pretty much the reason why Python projects are built into wheels and published on PyPI. 

This is not critical to get merged, only if that's appropriate given the current circumstances. Many thanks!

alanfranz:
> Thanks for looking at this! I don't have a concrete use case yet, but being able to bootstrap an environment from `requirements.txt` may be helpful (for me and others) instead of cloning the GitHub repository. That's pretty much the reason why Python projects are built into wheels and published on PyPI.

You can use github source in requirements.txt as well; but, myhoard is not a library. You should use it as an application; so, installing it from requirements in another virtualenv sound strange to me.

Technically, I don't think it's a real problem, it's more a matter of policy than anything else.

alexey-tereshenkov-aiven:
> > Thanks for looking at this! I don't have a concrete use case yet, but being able to bootstrap an environment from `requirements.txt` may be helpful (for me and others) instead of cloning the GitHub repository. That's pretty much the reason why Python projects are built into wheels and published on PyPI.
> 
> You can use github source in requirements.txt as well; but, myhoard is not a library. You should use it as an application; so, installing it from requirements in another virtualenv sound strange to me.

You may want to install your applications into a virtual environment as well e.g. with `pipx`. Totally agree everything is possible and installing from sources is a valid path, and having a wheel is not critical as I wrote before. It brings convenience and benefits for certain people and use cases and in general tidier than sources (you can query PyPI packages to see what is the Python version compatibility, have releases yanked, platform compatibility etc). 

But please merge only if you deem this appropriate!

> 
> Technically, I don't think it's a real problem, it's more a matter of policy than anything else.

.github/workflows/publish-pypi.yaml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        persist-credentials: false
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.10
```
st3fan, line 26: Is that not parsed into `3.1` ?

alexey-tereshenkov-aiven: Oh yes, very likely, indeed. Fixed!

jelmer:
Anything left blocking this?

alanfranz:
> Anything left blocking this?

1) We don't know whether we need or want this. This is not a library.
2) API token should be generated, then configured in github secrets.

PR #104, opened 2022-11-28
Nicois/sre 5264 relaxed timeout during optimize

nicois:
As part of the workaround for xtrabackup incompatibility with altered tables with the latest version of mysql, myhoard scans the list of table names and runs OPTIMIZE TABLE on each of them, prior to each basebackup.

Some mysql services have many table name, taking well over the default timeout to return that list. 

Modify the mysql cursor’s timeout during this optimise operation, without altering other timeouts.

Additional minor changes are made to stream back the list of tables to myhoard, and to use a simpler variable structure.


packi (approved):
LGTM

PR #105, opened 2022-11-28
Make running optimize for MySQL 8.0.30 optional

kmichel-aiven:
This is disabled by default and can be enabled with a new option: `optimize_tables_before_backup`.

The operation can be slow on very large databases and is rarely needed, default to making it optional.


packi (approved):
LGTM

PR #107, opened 2022-12-07
Fix tests by switching GitHub actions back to Ubuntu 20.04.

jelmer:
As of recently, "ubuntu-latest" has changed from being "ubuntu-20.04" to "ubuntu-22.04". However, no MySQL 8.0.26 packages are available for that version which breaks the GitHub actions. See https://github.com/actions/runner-images/issues/6399

Moving back to Ubuntu 22.04 fixes the GitHub action, allowing us to move to Ubuntu 22.04 or a mix of versions at our leisure.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


aris-aiven (approved):
The decision to make latest mean 20.04 is hard to read from downstream consumers in my opinion.

PR #111, opened 2022-12-12
Add more typing in myhoard.binlog_scanner

jelmer:
Drive-by fix as I try to understand the myhoard codebase.


kmichel-aiven (approved):
Long term I think we should move away from dicts and do more dataclasses, but this is already better than yesterday's code :)

PR #113, opened 2022-12-12
Add typing for "SHOW SLAVE STATUS"

jelmer:
This helps indicate use of typoed fields or
code that fails to deal with "SHOW SLAVE STATUS" not returning any rows.

It also highlights the bug beneath the fix in 9629527a10d6d702d1a73c693db3e1e971d1bd61

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -10,6 +10,7 @@ from .util import (
     build_gtid_ranges,
     change_master_to,
     ERR_TIMEOUT,
+    get_slave_status,
     GtidRangeDict,
     make_gtid_range_string,
     mysql_cursor,
@@ -18,6 +19,7 @@ from .util import (
     read_gtids_from_log,
     relay_log_name,
     rsa_decrypt_bytes,
+    SlaveStatus,
     sort_and_filter_binlogs,
     track_rate,
 )
@@ -368,7 +370,7 @@ class RestoreCoordinator(threading.Thread):
         finally:
             self.basebackup_restore_operation = None
 
-    def refresh_binlogs(self):
+    def refresh_binlogs(self) -> None:
         self._fetch_more_binlog_infos(force=True)
         if not self.pending_binlogs:
             self.log.info("No binary logs available, marking restore completed immediately")
@@ -376,7 +378,7 @@ class RestoreCoordinator(threading.Thread):
         else:
             self.update_state(phase=self.Phase.applying_binlogs)
 
-    def apply_binlogs(self):
+    def apply_binlogs(self) -> None:
         binlogs = self._get_binlogs_to_apply()
         if not binlogs:
             return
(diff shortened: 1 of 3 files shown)


myhoard/restore_coordinator.py
```diff
@@ -1187,13 +1188,13 @@ def _start_sql_thread_if_not_running(self, slave_status):
...
-    def _check_sql_slave_status(self):
+    def _check_sql_slave_status(self) -> tuple[bool, int]:
         expected_range = self.state["current_executed_gtid_target"]
         expected_index = self.state["current_relay_log_target"]
 
         with self._mysql_cursor() as cursor:
-            cursor.execute("SHOW SLAVE STATUS")
-            slave_status = cursor.fetchone()
+            slave_status = get_slave_status(cursor)
+            assert slave_status
```
kmichel-aiven, line 1197: I think a "real" check and exception would be a better fit than an assert here.

Auto-recovering from that odd case is out of the scope of the PR, but something like this could be more ops-friendly if that happens:
```suggestion
            if not slave_status:
                raise Exception("No status available from SHOW SLAVE STATUS, perhaps MySQL was restarted while applying binlog.")
```

myhoard/restore_coordinator.py
```diff
@@ -1263,15 +1264,16 @@ def _check_sql_slave_status(self):
...
                 # Current file could've been updated since we checked it the first time before slave was stopped.
                 # Get the latest value here so that we're sure to start from correct index
-                cursor.execute("SHOW SLAVE STATUS")
-                current_file = cursor.fetchone()["Relay_Log_File"]
-                last_index = int(current_file.rsplit(".", 1)[-1])
-                if last_index > current_index:
-                    self.log.info("Relay index incremented from %s to %s after STOP SLAVE", current_index, last_index)
-                    current_index = last_index
+                replica_status = get_slave_status(cursor)
+                if replica_status:
```
kmichel-aiven, line 1268: The only case where this could be falsy without triggering the assert/exception above is if MySQL was restarted and forgot completely about its replication state, so something is odd and the state of the restoration is uncertain. I think we should fail here too and not just skip some part of the code.

myhoard/util.py
```diff
@@ -569,3 +569,15 @@ def run(self):
...
+
+class SlaveStatus(TypedDict):
+
+    Relay_Log_File: str
+    Slave_SQL_Running: str
+    Slave_SQL_Running_State: str
+
+
+def get_slave_status(cursor) -> Optional[SlaveStatus]:
+    cursor.execute("SHOW SLAVE STATUS")
```
kmichel-aiven, line 582: Can you add a comment about why this can be None here ? (ie. it's not just the typing of cursor that is inconvenient for us, it's a thing that happens for real when the replication was never configured).

kmichel-aiven (dismissed):

myhoard/util.py
jelmer: Done

myhoard/restore_coordinator.py
jelmer: Done

packi (approved):

PR #114, opened 2022-12-13
Add typing for StateManager and AppendOnlyStateManager

jelmer:


jelmer:
> Do we really need/want the generic type if it's used with a single type? @jelmer

There's 4 uses of StateManager and 3 of AppendOnlyStateManager in myhoard, and at least one of use of StateManager in aiven-core (each with a different type). I haven't updated those yet, since none of them have their state as a TypedDict yet and I didn't want this PR to become huge. 

Let me know if you'd prefer I include those changes here as well, rather than as bite-size followup PRs.

alanfranz (approved):
LGTM

jelmer:
> I suppose mypy is lenient when using a raw type when a generic was expected, then, right?

Yep, at least with the mode in which we're currently running it

PR #115, opened 2022-12-13
Add typing for gtid ranges and gtid executed

jelmer:


myhoard/binlog_scanner.py
```diff
@@ -14,7 +14,7 @@ class BinlogInfo(TypedDict):
     file_name: str
     file_size: int
     full_name: str
-    gtid_ranges: List[Tuple[int, int, str, int, int]]
+    gtid_ranges: List[GtidRangeDict]
```
jelmer, line 17: This is indeed changing the type, but the code in binlog_scanner actually assumes it's a dict.

alanfranz (approved):
LGTM

PR #117, opened 2022-12-16
Track when the SQL thread is restarted unexpectedly

jelmer:
When a SQL thread has died and is revived:

* log the last reason for why the SQL thread died
* count the number of restarts in a metric, along with the errno

### DIFF
--- a/myhoard/util.py
+++ b/myhoard/util.py
@@ -617,3 +617,17 @@ def get_slave_status(cursor) -> Optional[SlaveStatus]:
     """
     cursor.execute("SHOW SLAVE STATUS")
     return cursor.fetchone()
+
+
+def restart_unexpected_dead_sql_thread(cursor, slave_status, stats, log):
+    if slave_status["Last_SQL_Error"]:
+        log.warning(
+            "SQL Thread died, running 'START SLAVE SQL_THREAD'. Last reported error at %s: %d: %s",
+            slave_status["Last_SQL_Error_Timestamp"],
+            slave_status["Last_SQL_Errno"],
+            slave_status["Last_SQL_Error"],
+        )
+    else:
+        log.warning("Expected SQL thread to be running but it isn't. Running 'START SLAVE SQL_THREAD'")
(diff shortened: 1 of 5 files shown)


myhoard/util.py
```diff
@@ -571,3 +571,18 @@ def run(self):
                 self.stats.unexpected_exception(ex=exception, where="RateTracker.run")
                 self.stats.increase("myhoard.ratetracker.errors")
                 self._reset()
+
+
+def restart_unexpected_dead_sql_thread(cursor, slave_status, stats, log):
+    assert slave_status["Slave_SQL_Running"] != "Yes"
```
kmichel-aiven, line 577: I don't think we should have this assert here.
Normally, if we read the code sequentially and assume no concurrent operations, it will always be true and harmless.
But if the thread is restarted externally between the ifs that guard the calls to  `restart_unexpected_dead_sql_thread` and this line, then we will get an `AssertionError`.
This error will not really enforce any invariant but it will be one more cause of failure, we lose more than we gain by failing here.

test/test_util.py
```diff
@@ -280,3 +280,33 @@ def test_detect_running_process_id(self, cmd_str):
             raise AssertionError(f"Could not match command or matched twice:\n{output_bytes.decode('ascii')}")
         no_id, output_bytes = myhoard_util.detect_running_process_id("certainlynosuchprocesscurrentlyrunning")
         assert no_id is None
+
+
+def test_restart_unexpected_dead_sql_thread():
```
kmichel-aiven, line 285: nit. Add return type of `-> None`, our mypy config isn't great yet, but this will help detect issues as we increase the scope of mypy checks.

kmichel-aiven (dismissed):

myhoard/util.py
jelmer: I agree that normally this should always evaluate to true - but that should make it a great expression for an assert :)

Either way, I don't feel strongly about keeping it so removed.

test/test_util.py
jelmer: Done.

packi (dismissed):
LGTM but conflicts need to be resolved before merging.

packi (approved):
LGTM

PR #120, opened 2022-12-30
Print a better error message when the expected backup site is missing

jelmer:
This seems slightly better than a plain KeyError. Is there something better we can do in this situation?

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2019 Aiven, Helsinki, Finland. https://aiven.io/
 from .backup_stream import BackupStream
 from .binlog_scanner import BinlogScanner
-from .errors import BadRequest
+from .errors import BadRequest, UnknownBackupSite
 from .restore_coordinator import RestoreCoordinator
 from .state_manager import StateManager
 from .util import (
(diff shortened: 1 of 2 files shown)


myhoard/controller.py
```diff
@@ -669,7 +669,12 @@ def _binlog_uploaded(self, *, local_index, remote_key, stream):
 
     def _build_backup_stream(self, backup):
         stream_id = backup["stream_id"]
-        backup_site = self.backup_sites[backup["site"]]
+        backup_site_name = backup["site"]
+        try:
+            backup_site = self.backup_sites[backup_site_name]
+        except KeyError:
+            self.stats.increase("myhoard.unknown_backup_site", tags={"backup_site": backup_site_name})
+            raise AssertionError(f"backup site {backup_site_name} unknown (sites: {list(self.backup_sites.keys())!r})")
```
packi, line 677: I would change this to a specific exception so if this ends up in a log it's clear from the exception what could happen. Helps with classifying errors without stats.

myhoard/controller.py
```diff
@@ -774,7 +779,12 @@ def _create_restore_coordinator_if_missing(self):
             return
 
         options = self.state["restore_options"]
-        backup_site = self.backup_sites[options["site"]]
+        backup_site_name = options["site"]
+        try:
+            backup_site = self.backup_sites[backup_site_name]
+        except KeyError:
+            self.stats.increase("myhoard.unknown_backup_site", tags={"backup_site": backup_site_name})
+            raise AssertionError(f"backup site {backup_site_name} unknown (sites: {list(self.backup_sites.keys())!r})")
```
packi, line 787: Same here.

packi (dismissed):
NIT: I would probably extract the lookup into a function so we're sure that the exception and stats match in the error case.

myhoard/controller.py
jelmer: Done.

myhoard/controller.py
jelmer: Done.

packi (approved):

PR #121, opened 2023-01-05
Fix for failing backup when GTID executed line is too long [BF-1449]

alexole:

When a base backup is finished we parse the output of `xtrabackup` to get the information about the binary logs wrt to this backup, this information is used by backup streams to check if binlog is needed for the backup or could be skipped.

If this information is missing, there might be side effects, such as e.g. a new backup stream might think that a pending binlog (which is about to be purged) is needed by a new backup, even though a new backup binlog position is newer than this binlog.
So the problem is that we expect that the output line of `xtrabackup` always includes the complete GTID executed line, that's not always the case. The max length of log line in `xtrabackup` is 8192, thus it will truncate lines with lots of GTID executed entries to this length, which causes parsing issues and binlog info missing.
    
The fix is to get rid of parsing binlog information from the log, instead - fetch information from extra meta file, which should be more reliable.

### DIFF
--- a/myhoard/basebackup_operation.py
+++ b/myhoard/basebackup_operation.py
@@ -31,7 +31,7 @@ class BasebackupOperation:
 
     current_file_re = re.compile(r" Compressing, encrypting and streaming (.*?)( to <STDOUT>)?( up to position \d+)?$")
     binlog_info_re = re.compile(
-        r"MySQL binlog position: filename '(?P<fn>.+)', position '(?P<pos>\d+)'(, GTID of the last change '(?P<gtid>.+)')?$"
+        r"binlog_pos = filename '(?P<fn>.+)', position '(?P<pos>\d+)'(, GTID of the last change '(?P<gtid>.+)')?$"
     )
     lsn_re = re.compile(r" Transaction log of lsn \((\d+)\) to \((\d+)\) was copied.$")
     progress_file_re = re.compile(r"(ib_buffer_pool$)|(ibdata\d+$)|(.*\.CSM$)|(.*\.CSV$)|(.*\.ibd$)|(.*\.sdi$)|(undo_\d+$)")
@@ -59,6 +59,7 @@ class BasebackupOperation:
         self.encryption_algorithm = encryption_algorithm
         self.encryption_key = encryption_key
         self.log = logging.getLogger(self.__class__.__name__)
+        self.lsn_dir = None
         self.lsn_info = None
         self.mysql_client_params = mysql_client_params
         with open(mysql_config_file_name, "r") as config:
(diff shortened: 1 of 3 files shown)


myhoard/basebackup_operation.py
```diff
@@ -315,16 +323,25 @@ def _process_output_line_file_finished(self, line):
...
-        match = self.binlog_info_re.search(line)
-        if match:
-            self.binlog_info = {
-                "file_name": match.group("fn"),
-                "file_position": int(match.group("pos")),
-                "gtid": match.group("gtid"),
-            }
-            self.log.info("binlog info: %r", self.binlog_info)
-        return match
+    def _process_binlog_info(self):
```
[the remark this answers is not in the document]
alexole: This processes binlog info, not just some other info, there is much more in that file but we are not interested in it.

myhoard/basebackup_operation.py
```diff
@@ -315,16 +323,25 @@ def _process_output_line_file_finished(self, line):
...
+                    if match:
+                        self.binlog_info = {
+                            "file_name": match.group("fn"),
+                            "file_position": int(match.group("pos")),
+                            "gtid": match.group("gtid"),
+                        }
+                        self.log.info("binlog info: %r", self.binlog_info)
+                        break
+        except FileNotFoundError:
+            pass
```
rikonen, line 344: This error should probably not be ignored as we'll end up missing critical info if it fails.

myhoard/basebackup_operation.py
```diff
@@ -315,16 +323,25 @@ def _process_output_line_file_finished(self, line):
...
+                for line in binlog_file.readlines():
+                    match = self.binlog_info_re.search(line)
+                    if match:
+                        self.binlog_info = {
+                            "file_name": match.group("fn"),
+                            "file_position": int(match.group("pos")),
+                            "gtid": match.group("gtid"),
+                        }
+                        self.log.info("binlog info: %r", self.binlog_info)
+                        break
```
rikonen, line 342: The `for` loop should presumably have an `else` branch that raises an error in case we could not find the expected information.

rikonen (dismissed):
Binlog info is required information so failure to get it should not be ignored, otherwise looks good to me.

alexole:
@rikonen the old logic assumes it might be empty, therefore I did not change this part
https://github.com/aiven/myhoard/blob/aafd62e913bd748dadc5aee57d9e9e4a8778d707/myhoard/backup_stream.py#L870
OK, will change that.

alexole:
As agreed I added a warning message in case if binlog info is missing in the info file (this is possible when binary log is off for example), but I made `lsn_dir` and info file mandatory and calling now this method on successful backup only. @rikonen

rikonen (approved):
Looks good to me

PR #123, opened 2023-01-06
Add specific exception for abort requests

jelmer:


myhoard/basebackup_operation.py
```diff
@@ -21,6 +21,14 @@
 CURSOR_TIMEOUT_DURING_OPTIMIZE: int = 120
 
 
+class AbortRequested(Exception):
+    """The base backup operation was aborted by request."""
+
+    def __init__(self, reason):
```
kmichel-aiven, line 27: Missing parameter and return type hints.

myhoard/basebackup_operation.py
```diff
@@ -21,6 +21,14 @@
...
+        super().__init__()
```
kmichel-aiven, line 28: The exception would be more informative in logs and traces if a message is passed to the base:
```suggestion
        super().__init__(f"reason: {abort_reason}")
```

This turns `print(error)` into `AbortRequested("reason: something")` instead of the mysterious `AbortRequested()`.

kmichel-aiven (dismissed):

kmichel-aiven (dismissed):
I think there was a push bug somewhere, the code is the same as before :D

jelmer:
> I think there was a push bug somewhere, the code is the same as before :D

Sorry, not sure what happened there! Should be good now.

kmichel-aiven (approved):

PR #128, opened 2023-01-09
backup_stream: Add typing

jelmer:


aris-aiven (dismissed):
For imports that are only used for `mypy`, would you mind guarding them with a `if TYPE_CHECKING` ?

aris-aiven (approved):
LGTM, thank you for the typing improvements!

PR #131, opened 2023-01-25
Provide a single metric to track basebackup restoration

nicois:
While there are a few metrics which are exposed to assist in tracking individual parts of a restore, none of them are sufficient on their own to track progress.

By creating a helper class to track progress from disparate sources, it is possible to provide another metric which monotonically increases throughout the basebackup restore process, including while replaying binlogs.

### DIFF
--- a/myhoard/basebackup_restore_operation.py
+++ b/myhoard/basebackup_restore_operation.py
@@ -253,7 +253,12 @@ class BasebackupRestoreOperation:
         assert self.proc is not None
         increase_pipe_capacity(self.proc.stdin, self.proc.stdout, self.proc.stderr)
 
-        sender_thread = InputSenderThread(stats=self.stats, stream_handler=self.stream_handler, stream=self.proc.stdin)
+        sender_thread = InputSenderThread(
+            stats=self.stats,
+            stream_handler=self.stream_handler,
+            stream=self.proc.stdin,
+            percent_complete_callback=self.progress.update,
+        )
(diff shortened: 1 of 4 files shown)


rikonen:
Is there some internal ticket reference explaining which problem we're trying to solve with this? Answer to what problem we're trying to solve may affect the answer to this but: Since the duration of basebackup restoration compared to binlog restoration is completely unknown and the binlogs to restore may be a moving target the percentage value this provides is probably wildly inaccurate in many (most?) cases so reporting progress as percentage might be somewhat confusing and just monotonically increasing value could work to show something is happening.

nicois:
@rikonen here is a brief example of the benefits of this metric: https://aiven.atlassian.net/browse/SRE-5536

I have developed this approach to deal with the fact that the exact duration of each phase is unknown, and in the case of binlogs that there may be multiple passes, while still providing a monotonic metric which starts at 0 and ends at 100.

While this cannot be used to calculate time remaining, at the very least it will more easily allow us to be alerted when a restoration is failing as opposed to taking a long time, either because the basebackup is large or because the replaying of binlogs is taking a long time.

rikonen:
Added comment to the internal ticket but essentially wondering if the existing basebackup progress reporting could just be tweaked a bit to fix existing metrics not showing any progress when basebackup is being prepared, which afaik is the only practical issue we have at the moment that this would address and change would be much simpler.

PR #132, opened 2023-01-25
README: Add a list of metrics exported by myhoard

jelmer:
Ideally these would all be documented, but I didn't really want to spend the time to do that - might come back to that alter. Having the list seems at least better than having to grep the code, especially for potential outside users.


aris-aiven (approved):
Thanks for compiling the list

PR #133, opened 2023-01-27
Report progressing of basebackup

yaolu-aiven:

When xtrabackup is working, it is constantly emitting logs similar to
  `2023-01-27T00:28:00.109513-00:00 1 [Note] [MY-011825] [Xtrabackup] >> log scanned up to (9125452372)`
every second.

If xtrabackup is still emitting logs, it can be considered the base backup is still progressing, and thus can be utilized in alerting suppression on a super large database.

(Originated from Aiven internal ticket SRE-5535)

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -908,6 +908,7 @@ class BackupStream(threading.Thread):
             # use longer than normal timeout here with multiple retries and increasing timeout.
             connect_params = dict(self.mysql_client_params)
             for retry, multiplier in [(True, 1), (True, 2), (False, 3)]:
+                self.stats.gauge_int("myhoard.basebackup.progressing", 1)
                 try:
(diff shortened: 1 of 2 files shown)


rikonen:
While this can be used to solve the problem this will require some new instrumentation on our side, plus this makes it more difficult to spot cases where it really does get more or less permanently stuck in the prepare phase as we still want to react at some point. I suggested a bit different approach in another internal ticket SRE-5536.

To recap the alternative suggestion here: Make the basebackup restore progress up to 99.0% in the downloading phase. After we've reached that then start looking for log lines matching the `log scanned up to ...` pattern and when we see such line and at least 60 seconds has passed since last progress update and we're below 99.99% progress, increase current progress by 0.01%. So when download finishes we're at 99.00%, one minute later if this keeps on getting logged we're at 99.01%, and 100 minutes later we're at 99.99% and there's no further progress until the preparation actually finishes so something ends up escalating and we don't need to change anything for components making use of current progress reporting.

PR #135, opened 2023-02-02
provide consistent progress report for basebackup

yaolu-aiven:

When xtrabackup is working, it is constantly emitting logs similar to
  `2023-01-27T00:28:00.109513-00:00 1 [Note] [MY-011825] [Xtrabackup] >> log scanned up to (9125452372)`
every second.

The change will allow myhoard to emit progress at 0.01% incremental at 30 seconds fixed interval, facilitating monitoring of backup progressing/progress.

My thought on potential edge cases.

* Is there any chance of backtracking progress?

Yes, if the database is really gigantic and has similar table sizes for different tables.
e.g. in the extreme case here, assume a database has 10 tables of equal size,
In this case, the 0.01% incremental per 30 seconds would take 10% / 0.01% * 30 = 30000 seconds to cause backtracking of progress. It's possible but it is an edge case. If there is any report of it, the incremental frequency can be updated to be configurable.

(Originated from Aiven internal ticket SRE-5535, testing logs attached in the ticket)

### DIFF
--- a/myhoard/basebackup_operation.py
+++ b/myhoard/basebackup_operation.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2019 Aiven, Helsinki, Finland. https://aiven.io/
 from contextlib import suppress
+from datetime import datetime, timedelta
 from distutils.version import LooseVersion  # pylint:disable=deprecated-module
 from myhoard.errors import XtraBackupError
 from myhoard.util import get_mysql_version, mysql_cursor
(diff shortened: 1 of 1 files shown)


yaolu-aiven:
The failure seems to be not related to the code change. Will have a look separately.

rikonen (changes requested):
This should probably work a bit differently. (Btw when I was commenting on something related to this earlier I was referring to the case of restoring basebackup and the prepare phase of that so my comment might not have made sense in the context of creating basebackup.)

Currently we always increment progress by 0.01 every 30 seconds even if progress has otherwise been incremented. The extra increment can cause progress to go beyond 100% and progress can go downwards when processing a file finishes.

I'd rather have something like this:

1. Whenever file upload completes update `self.current_progress_update_time` so that we do not do the extra updates when we have natural progress due to completion of individual files
2. Only increase the virtual progress if we actually see the log line mentioned in commit message, not for any log line (and only when more than 30s has elapsed since last update)
3. Instead of adding `self.current_progress_incremental_delta` to `self.current_progress` when handling the log line, increment `self.current_progress_incremental_delta` by 0.01 and have `_update_progress` always add `self.current_progres_incremental_delta` to the estimated progress (with cap to max 100.0% for total progress). That way progress will never go down.

alanfranz:
Yao doesn't work for Aiven anymore. I think we can close this, indeed.

PR #136, opened 2023-02-03
Support backup intervals greater than 24 hours

kathia-barahona:
Enables backup scheduling for intervals greater than 24 hours. This mainly changes the way we calculate the `_current_normalized_backup_timestamp` by considering the previous normalized backup timestamp.

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -17,5 +17,6 @@ pytest-mock
 pytest-timeout
 pytest-xdist
 responses
+time-machine
 types-PyMySQL
 types-requests
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -851,6 +851,17 @@ class Controller(threading.Thread):
         else:
             self.log.info("Newly created restore coordinator is already in completed state")
 
+    def _previous_normalized_backup_timestamp(self) -> Optional[str]:
+        normalized_backup_times = [
+            stream.state["normalized_backup_time"]
+            for stream in self.backup_streams
+            if stream.state["normalized_backup_time"]
+        ]
+
+        if not normalized_backup_times:
+            return None
+        return max(normalized_backup_times)
+
     def _current_normalized_backup_timestamp(self) -> str:
         """Returns the closest historical backup time that current time matches to (or current time if it matches).
         E.g. if backup hour is 13, backup minute is 50, current time is 15:40 and backup interval is 60 minutes,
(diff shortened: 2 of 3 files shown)


requirements.dev.txt
```diff
@@ -17,5 +17,6 @@ pytest-mock
 pytest-timeout
 pytest-xdist
 responses
+time-machine
```
kathia-barahona, line 20: **comment** using this module just for mocking time and test default intervals.

alanfranz (dismissed):
I don't understand how the ```backup_disabled_until``` value should be configured. We're disabling backups for X minutes... from when? From when the option is set? But then, we never unset this. If I find this config set, when should I expect the new backup to happen?

I think the "backup_disabled_until" should be a full timestamp (I'm not sure what format we're using elsewhere, take a look around). If the current time when trying to take a backup is before that, just skip that backup.

kathia-barahona:
> I don't understand how the `backup_disabled_until` value should be configured. We're disabling backups for X minutes... from when? From when the option is set? But then, we never unset this. If I find this config set, when should I expect the new backup to happen?
> 
> I think the "backup_disabled_until" should be a full timestamp (I'm not sure what format we're using elsewhere, take a look around). If the current time when trying to take a backup is before that, just skip that backup.

Yes I see your point, tho I thought the functionality was more about increasing the time interval between backups, because currently is at least `backup_interval_minutes / 2`.  Maybe I got confused when talking with @alexole. @alexole is it better to just change it with a timestamp?

alanfranz:
You can implement it as you see fit, but we need to keep the naming clean. If you just want to be able to override the backup_interval... you could call this option ```backup_interval_override_seconds```. And: please make sure such override is actually respected (we had a bug with backup_interval that made it impossible to span more than 24h) & validated.

alanfranz:
@kathia-barahona remember to re-add Team Bruteforce (and or team OSS in this case) as reviewers, otherwise you can get stuck at waiting for a single person (me, in this scenario :-) )

alanfranz:
I'm not sure I understand the change, but the backup logic in myhoard was pretty complex. If anybody wants to chime in in place of me, please do so; otherwise I'll take another look in the next days.

rikonen:
Pull request descriptions talks about feature that this pr no longer implements and first commit has code that is deleted in the second commit. Should squash commits and update commit and pr description to match what this actually does, now this pr looks quite confusing.

kathia-barahona:
> Pull request descriptions talks about feature that this pr no longer implements and first commit has code that is deleted in the second commit. Should squash commits and update commit and pr description to match what this actually does, now this pr looks quite confusing.

oops, you are right. will update this

Ormod:
The change is in conflict with master.

kmichel-aiven (approved):

PR #140, opened 2023-02-08
Allow rebuilding tables before restoring binlogs

kmichel-aiven:

This can avoid data corruption caused by multiple bugs in MySQL 8.0.30+ when updating from an earlier version.

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -6,6 +6,7 @@ from .binlog_downloader import download_binlog
 from .errors import BadRequest
 from .state_manager import StateManager
 from .statsd import StatsClient
+from .table import Table
 from .util import (
     add_gtid_ranges_to_executed_set,
     build_gtid_ranges,
@@ -99,6 +100,7 @@ class RestoreCoordinator(threading.Thread):
         getting_backup_info = "getting_backup_info"
         initiating_binlog_downloads = "initiating_binlog_downloads"
         restoring_basebackup = "restoring_basebackup"
+        rebuilding_tables = "rebuilding_tables"
         refreshing_binlogs = "refreshing_binlogs"
         applying_binlogs = "applying_binlogs"
         waiting_for_apply_to_finish = "waiting_for_apply_to_finish"
@@ -135,6 +137,7 @@ class RestoreCoordinator(threading.Thread):
         last_poll: Optional[float]
         last_processed_index: Optional[int]
         last_renamed_index: int
+        last_rebuilt_table: Optional[Tuple[str, str]]
         mysql_params: Optional[Dict[str, Any]]
         phase: "RestoreCoordinator.Phase"
         prefetched_binlogs: Dict
@@ -160,10 +163,12 @@ class RestoreCoordinator(threading.Thread):
         mysql_relay_log_index_file,
         mysql_relay_log_prefix,
         pending_binlogs_state_file,
+        rebuild_tables: bool,
         restart_mysqld_callback,
         rsa_private_key_pem,
         site: str,
         state_file: str,
+        state_manager_class: type[StateManager["RestoreCoordinator.State"]] = StateManager,
         stats: StatsClient,
         stream_id: str,
         target_time=None,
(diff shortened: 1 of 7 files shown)


myhoard/web_server.py
```diff
@@ -133,6 +133,8 @@ async def status_update(self, request):
             elif body_mode == Controller.Mode.observe:
                 self.controller.switch_to_observe_mode()
             elif body_mode == Controller.Mode.restore:
+                if not isinstance(body.get("rebuild_tables"), (bool, type(None))):
```
alexole, line 136: If I am reading this correctly, in case of body containing:
```
rebuild_tables: null
```
It will pass None to the restore_backup which is not expected, as it's not optional, even though I assume the code will work.

alexole, line 136: I think the validation is not 100% correct, if the body contains:
```
"rebuild_tables": null
```
the validation will succeed and later the None value will be passed over instead of False.

myhoard/restore_coordinator.py
```diff
@@ -430,6 +436,66 @@ def restore_basebackup(self) -> None:
...
+                        # normal tables (like mysql.innodb_index_stats and mysql.innodb_table_stats).
+                        # The known ones are already skipped but if we encounter more, we just skip them.
+                        self.log.error(
+                            "Could not rebuild %s.%s, error: %s, skipping it",
+                            escaped_table_schema,
+                            escaped_table_name,
+                            str(e),
+                        )
+                    else:
+                        raise
```
alexole, line 494: If `myhoard` gets restarted I think it would be nice to skip errors which will happen on the table which is running ALTER TABLE already. (needs to be checked what happens in this case and if we need to actually wait for this to finish).

Also, this I am not sure myself, but maybe we should not hard stop in case of any problem with rebuilding, we don't know yet other cases and if we roll this out to many services, we might not be able to genbump due to that.
Maybe we should try to repair and instead of failing collect a metric and have alerts for those. What do you think?

myhoard/restore_coordinator.py
```diff
@@ -430,6 +436,66 @@ def restore_basebackup(self) -> None:
...
+                        ("mysql", "innodb_index_stats"),
+                        ("mysql", "innodb_table_stats"),
+                    }
+                ],
+                key=lambda table: (table["TABLE_SCHEMA"], table["TABLE_NAME"]),
+            )
+            estimated_total_bytes = sum(table["TABLE_ROWS"] * table["AVG_ROW_LENGTH"] for table in tables)
+            estimated_progress_bytes = 0
+            self.log.info("Will rebuild %s tables, estimated total size: %s bytes", len(tables), estimated_total_bytes)
+            for table_index, table in enumerate(tables, start=1):
```
rikonen, line 466: Might be good to split this processing so that there's separate step for listing the tables that need to be rebuilt and those are stored in state as well as the latest table that has been built thus far so that operation resumes from where it was left if it's aborted instead of starting over from beginning.

myhoard/web_server.py
kmichel-aiven: Ah, yes. Will fix that.

myhoard/restore_coordinator.py
kmichel-aiven: The main expected cause of failure would be a corrupted table, I'd rather have the genbump fail hard in that case, while there is still a previous node running.

If it doesn't, then we restore a possibly corrupted database, and then remove the previous node.
Nobody will look at the metrics until it's too late and corruption becomes visible.

For the restart, I'll check how MySQL behaves.

myhoard/restore_coordinator.py
kmichel-aiven: I didn't really need to save the table list to be able to resume.

Do you think about storing the list of tables for some specific reason or just as a possible implementation ?

myhoard/restore_coordinator.py
kmichel-aiven: I tried concurrent ALTER and MySQL does not raise errors, it just does both.

myhoard/restore_coordinator.py
rikonen: Current implementation should be fine. I was thinking of not rerunning the query to list tables but refreshing those might be good idea and some pathological services could have extremely large number of tables so avoiding storing those in state makes sense.

rikonen (approved):
Looks good to me. Leaving open if @alexole still wanted to have a look

alexole (approved):

PR #141, opened 2023-02-09
controller: don't request new backup if we're already at the maximum

aiven-amartin:

Currently if we get a bunch of backups for whatever reason, this wave pattern is amplified by the way we request backups. 

This PR attempts to fix that by not requesting a scheduled backup if we're already over the maximum count, which should short-circuit this wave.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -81,14 +81,6 @@ class RestoreOptions(TypedDict):
     target_time_approximate_ok: bool
 
 
-def sort_completed_backups(backups: List[Backup]) -> List[Backup]:
-    def key(backup):
-        assert backup["completed_at"] is not None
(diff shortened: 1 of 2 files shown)


jelmer:
Am I understanding correctly that this would only prevent scheduling new backups from myhoard's internal scheduling loop, not external requests? That seems like a feature, but perhaps good to emphasize here.

rikonen:
I would perhaps wait and see how the cloud provider fixes progress since at the moment it looks like we cannot rely on the object storage API and it may not be meaningful to try to work around the most critical component the software deals with being broken.

aiven-amartin:
Closing for now in favour of investigation into how to handle unstable object storage APIs as the real root cause.

PR #142, opened 2023-02-10
Reupload stream closed/completed marker in case of state mismatch

rikonen:
Sometimes we end up in a situation where the closed.json/completed.json for a stream does not exist in object storage but local state says the stream has been closed/completed. This condition should not be possible because object storage is updated before updating local state and any error in updating object storage should cause the local state not to be updated.

To recover from this supposedly impossible situation, have controller pass the closed/completed timestamp read from object storage when it creates a backup stream object and if backup stream sees that the object storage closed/completed marker is not present even though it should be then it uploads it again, hopefully resolving the issue.

Fixes (or works around) invalid state that the service can sometimes end up in.


alexole (approved):
Looks reasonable

rikonen:
This is not the correct fix for the problem. Apparently the files are uploaded successfully but the cloud provider API is currently broken and can return incomplete results. Uploading again does not fix the situation, the files can be missing from the responses no matter how hard you try uploading them. The last commit here would probably be valid, though, but closing the entire pr for now.

PR #143, opened 2023-02-13
support restore_basebackup_max_target_time

kathia-barahona:
Config parameter for limiting the basebackup selection. This way we can only allow restoring basebackups that were created before any incident.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,6 +342,11 @@ The full path and file name prefix of relay logs. This must be the same as the
 corresponding MySQL configuration option except full path is always required
 here.
 
+**restore_basebackup_max_target_time**
+
+Maximum target time allowed for basebackup restoration, meaning that only backups
+created before this time can be restored, otherwise the request will fail.
+
 **restore_free_memory_percentage**
 
 Maximum percentage of system memory to allow xtrabackup to use while
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -47,6 +47,7 @@ ERR_BACKUP_IN_PROGRESS = 4085
 
 
 class BaseBackup(TypedDict):
+    start_ts: float
     end_ts: float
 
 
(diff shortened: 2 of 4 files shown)


rikonen:
It's not quite clear what's the point of having a config option for this vs just not making requests with timestamps you don't want to restore.

kathia-barahona:
> It's not quite clear what's the point of having a config option for this vs just not making requests with timestamps you don't want to restore.

Yes, though the main idea is to allow us a way to avoid any requests that try to restore corrupted backups in case we got incidents. Maybe the naming is not the best. @kmichel-aiven  please correct me if I am wrong

rikonen:
For our internal use cases requests don't go to MyHoard directly so we can / probably should restrict these elsewhere

kmichel-aiven:
> It's not quite clear what's the point of having a config option for this vs just not making requests with timestamps you don't want to restore.

The idea is to restore to the same date as when using `target_time`, but starting from an older basebackup and replaying more binlogs to reach `target_time`. 
This will be slower but helpful if we want to avoid a more recent basebackup, if we suspect something is wrong with the basebackup but not the binlogs.

rikonen:
OK, that can make sense but now the code just raises an error if you give it newer timestamp instead of picking older basebackup to restore. (Also, possibly cleaner approach would be to add some separate operation that can be used to flag a backup as broken and those backups would then automatically be excluded by the existing logic for skipping broken backups.)

kathia-barahona:
> OK, that can make sense but now the code just raises an error if you give it newer timestamp instead of picking older basebackup to restore. (Also, possibly cleaner approach would be to add some separate operation that can be used to flag a backup as broken and those backups would then automatically be excluded by the existing logic for skipping broken backups.)

actually yes, tho that's when we get a restoring request with a specific stream id. I probably missed other cases we perform restores automatically. Need to take a look at this.

PR #145, opened 2023-02-17
mark basebackups as broken when restore fails

kathia-barahona:
Will close PR: https://github.com/aiven/myhoard/pull/143

Following @rikonen idea about marking basebackups as broken instead of having a target time for basebackups. Will create a ticket for marking/unmarking basebackups as broken via API request.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -54,6 +54,7 @@ class Backup(TypedDict):
     basebackup_info: BaseBackup
     closed_at: Optional[float]
     completed_at: Optional[float]
+    broken_at: Optional[float]
     recovery_site: bool
     stream_id: str
     resumable: bool
@@ -287,6 +288,10 @@ class Controller(threading.Thread):
                     continue
                 if not backup["basebackup_info"]:
                     raise ValueError(f"Backup {backup!r} cannot be restored")
+
+                if backup.get("broken_at"):
+                    raise ValueError(f"Cannot restore a broken backup: {backup!r}")
+
                 if target_time:
                     if target_time < backup["basebackup_info"]["end_ts"]:
                         raise ValueError(f"Requested target time {target_time} predates backup completion: {backup!r}")
@@ -559,6 +564,7 @@ class Controller(threading.Thread):
             for site_and_stream_id in streams:
                 basebackup_compressed_size = None
                 basebackup_info = {}
+                broken_info = {}
                 closed_info = {}
                 completed_info = {}
                 for info in file_storage.list_iter(site_and_stream_id):
@@ -573,6 +579,8 @@ class Controller(threading.Thread):
                             info_str, _ = file_storage.get_contents_to_string(info["name"])
                             basebackup_info = json.loads(info_str.decode("utf-8"))
                             seen_basebackup_infos[site_and_stream_id] = basebackup_info
+                    elif file_name == "broken.json":
+                        broken_info = parse_fs_metadata(info["metadata"])
                     elif file_name == "closed.json":
                         closed_info = parse_fs_metadata(info["metadata"])
                     elif file_name == "completed.json":
@@ -586,6 +594,7 @@ class Controller(threading.Thread):
                 backups.append(
                     {
                         "basebackup_info": basebackup_info,
+                        "broken_at": broken_info.get("broken_at"),
                         "closed_at": closed_info["closed_at"] if closed else None,
                         "completed_at": completed_info["completed_at"] if completed else None,
                         "recovery_site": site_config.get("recovery_only", False),
@@ -1140,6 +1149,7 @@ class Controller(threading.Thread):
         self._process_local_binlog_updates()
         self._extend_binlog_stream_list()
         if self.restore_coordinator.phase == RestoreCoordinator.Phase.failed_basebackup:
+            self._mark_failed_restore_backup_as_broken()
             self._switch_basebackup_if_possible()
         if self.state["promote_on_restore_completion"] and self.restore_coordinator.is_complete():
             self.state_manager.update_state(
@@ -1150,6 +1160,20 @@ class Controller(threading.Thread):
                 mode=self.Mode.promote,
                 restore_options={},
             )
+
+    def _mark_failed_restore_backup_as_broken(self) -> None:
+        broken_backup = None
+        failed_stream_id = self.state["restore_options"]["stream_id"]
+        backups = self.state["backups"]
+        for backup in backups:
+            if backup["stream_id"] == failed_stream_id:
+                broken_backup = backup
+                break
+
+        if not broken_backup:
+            raise Exception(f"Stream {failed_stream_id} to be marked as broken not found in completed backups: {backups}")
+
+        self._build_backup_stream(broken_backup).mark_as_broken()
 
     def _mark_periodic_backup_requested_if_interval_exceeded(self):
         normalized_backup_time = self._current_normalized_backup_timestamp()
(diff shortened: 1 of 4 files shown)


test/test_controller.py
```diff
@@ -1282,3 +1283,198 @@ def test_collect_binlogs_to_purge():
...
+            backups = m_controller.state["backups"]
+            assert len(backups) == expected_num
+            assert all(backup["completed_at"] is not None for backup in backups)
+
+        while_asserts(lambda: all_backups_completed(expected_num=2), timeout=30)
+
+        m_controller.stop()
+
+        # get the most recent backup (corrupted)
+        corrupted_backup = sort_completed_backups(m_controller.state["backups"])[1]
```
alanfranz, line 1329: this is not an actually corrupted backup, is it? it's the side effect monkeypatching down there that makes it corrupted.

test/test_controller.py
```diff
@@ -1282,3 +1283,198 @@ def test_collect_binlogs_to_purge():
     assert binlogs_to_purge == binlogs
     assert only_inapplicable_binlogs is False
     log.info.assert_any_call("Binlog %s has been replicated to all servers, purging", 3)
+
+
+@patch.object(RestoreCoordinator, "MAX_BASEBACKUP_ERRORS", 2)
+def test_restore_failed_basebackup(
```
alanfranz, line 1289: I'm a great nitpicker for naming; I don't understand what this should test.

test_basebackup_is_marked_as_failed_if_restore_fails ?

test/test_controller.py
```diff
@@ -1282,3 +1283,198 @@ def test_collect_binlogs_to_purge():
...
+            corrupted_backup = current_backups[1]
+            # failed basebackup should be marked as broken
+            assert corrupted_backup["broken_at"] is not None
+
+            # verify the empty mysql restored all data.
+            with mysql_cursor(**mysql_empty.connect_options) as cursor:
+                expected_row_count = data_generator.row_count
+                cursor.execute("SELECT COUNT(*) AS count FROM db1.t1")
+                result_row_count = cursor.fetchone()["count"]
+                assert expected_row_count == result_row_count
```
alanfranz, line 1379: this is another assertion, different from "broken backup is marked as such". 

how does this related to the other assertion? there're multiple backups, some are broken, some are not? We should make it clear what we're testing.

If the backup is broken, we should NOT be able to query the rows, right? OTC, if some restoration succeeded, then we should get the data. 

Maybe (that's just what I'm hypothesizing) you wanted to verify that, if the broken backup holds different data than the non-broken one, the non-broken one is the one that actually gets restored. But I don't see this kind of assertion, if I understand the code properly.

Also: the monkeypatching is far from ideal; the best thing would being able to make multiple backups, then corrupt one of them so that our code actally throws an exception.

Another path to explore: split tests. In one test, you just assert that, if a backup is marked as broken, its content is not restored (you can try marking it as broken "manually" somehow), by verifying what's the state of the db after the test.

In another test, you verify that, if an exception happens during restore, then such backup is marked as broken, there, you don't verify the final db content because you don't care.

test/test_controller.py
```diff
@@ -1282,3 +1283,198 @@ def test_collect_binlogs_to_purge():
...
+    finally:
+        m_controller.stop()
+        if new_controller:
+            new_controller.stop()
+
+
+@patch.object(BackupStream, "remove", autospec=True)
+def test_purge_old_backups(
```
alanfranz, line 1387: From this name I don't understand what the expected behaviour is :-(

myhoard/controller.py
```diff
@@ -1231,7 +1255,14 @@ def _purge_old_backups(self):
         # is called repeatedly so if there are for any reason more backups
         # to drop they will be dropped soon enough
         purgeable = sort_completed_backups(purgeable)
+        broken_backups = sum(backup["broken_at"] is not None for backup in purgeable)
```
alanfranz, line 1258: nit: broken_backups_count would be a better name

myhoard/controller.py
```diff
@@ -1231,7 +1255,14 @@ def _purge_old_backups(self):
...
         backup = purgeable[0]
+
+        if backup["broken_at"] is None and broken_backups and broken_backups == len(purgeable) - 1:
```
alanfranz, line 1261: I admit I'm not sure I understand this logic, and since I don't understand the test name, I'm left a bit in the dark. What is the outcome you desire? The idea is that if some backup is corrupted, but this one isn't, we don't delete it, right?

Backups are sorted left-to-right in decreasing age order (i.e. backups[-1] is the most recent, backups[0] is the oldest). When purging, we usually try to purge the oldest backup first (hence the purgeable[0]) above here.

IMHO (but verify, I may be missing some piece) the ideal logic would be:
- if there's any broken backup in the chain, purge the oldest broken backup and leave all non-broken backups there;
- if there're no broken backups, just purge the oldest.

alanfranz (dismissed):
Some questions on the specific files.

Also one more question: will this play nicely with pre-existing backups in prod where they don't have a broken_at key at all in metadata?

myhoard/controller.py
rikonen: > if there's any broken backup in the chain, purge the oldest broken backup

We store binlogs per basebackup and while the basebackup may be broken the binlogs associated with it are probably valid. Deleting a backup from middle, broken or no, would leave a hole in binlogs and break our ability to restore to any point in time since the oldest backup and we should not do that.

test/test_controller.py
kathia-barahona: I wouldn't say the test was intended only to check if the basebackup was marked. I realized that the restore/fail  flow was not entirely tested. So, the main idea is to verify if the selected backup fails, the controller is be capable of retrying with previous backup. Making sure the failed backup is been marked is just one part of it. 
 So that's why I went with a more "generalized" naming, do you have a better suggestion?

myhoard/controller.py
alanfranz: Thanks Rauli! OK, this makes sense to me (I was missing some piece after all) . So we should either 

a) not alter the purging logic at all; this is quite safe, and the restore will pick the correct basebackup.

or

b) try purging old, broken basebackups, if the oldest is broken, e.g.

if we have basebackups T1, T2, T3, T4, T5; with T1 being the oldest; if T1, T2 and T4 are broken, we can safely delete T1 and T2 (and all dependent binglogs) as well as delete them, since there's no way to restore them or anything that depends on them, while, OTC, we should retain T4 since we can still restore from T3 and use the T4 binlogs.

BUT:
I'd probably not enable this logic in production from the beginning; if we accidentally mark a basebackup as broken, we may end up deleting too many basebackups. Probably I'd add the logic with just a metric/logline, to see if it can really happen.

kathia-barahona:
>Also one more question: will this play nicely with pre-existing backups in prod where they don't have a broken_at key at all in metadata?

Shouldn't be a problem. We fill the metadata of each backup based on the jsons we store per status, so even if there is no `broken.json` for the backup, the metadata will still have the key. 
https://github.com/aiven/myhoard/blob/eb829289b3292963ac1b7211a75264ddc04c0e77/myhoard/controller.py#L567
https://github.com/aiven/myhoard/blob/eb829289b3292963ac1b7211a75264ddc04c0e77/myhoard/controller.py#L594-L605

test/test_controller.py
alanfranz: My suggestion would be to split tests (one behaviour asserted per test) and use names that assert the behaviour you'd like to test. I know that not all of our tests follow that convention, but it's just great for self-documentation and for people reading it in the future.

alanfranz:
> Shouldn't be a problem. We fill the metadata of each backup based on the jsons we store per status, so even if there is no `broken.json` for the backup, the metadata will still have the key.

A test for this would be great, btw!

test/test_controller.py
kathia-barahona: so it goes like this:

1. Attempt to restore a "corrupted" backup
2. Fail all attempts  (first 2 side effects)
3. RestoreController enters into "failed_basebackup" phase
4. Controller tries to restore previous backup
5. RestoreController retries with that backup (we assume this one is healthy) (no side effects here)
6. RestoreController successfully restores backup and enters into "completed" phase

So, that's why the 2 `wait_for_condition`: (1st one checks attempt with corrupted backup and 2nd with healthy backup)
https://github.com/aiven/myhoard/blob/eb829289b3292963ac1b7211a75264ddc04c0e77/test/test_controller.py#L1360-L1365


And yes, I did tried actually corrupting the xbstream file from the basebackup, but it was quite ugly. And IMHO this isn't the test that should cover it, rather do a separate test under `test_restore_coordinator`.

myhoard/controller.py
kathia-barahona: as discussed during daily, in case the oldest backup is the only non-broken, it will be only purged when a new backup is generated

kathia-barahona:
seems one of the tests is flaky :/ need to see how to fix it.

kmichel-aiven (approved):

PR #148, opened 2023-03-01
remove tests for percona-xtrabackup-8.0.28-20-01

kathia-barahona:
checks  for this version are currently failing, I guess skipping tests for  mysql 8.0.26 is alright

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -83,10 +83,6 @@ jobs:
       # skipping matrix items that aren't meaningful
       matrix:
         include:
-          - mysql-version: "8.0.26"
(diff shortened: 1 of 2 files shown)


packi (commented):
nit: I'd probably change that line to reference a newer version as well, if we're removing the testing for it:
https://github.com/aiven/myhoard/blob/master/README.md?plain=1#L791

kathia-barahona:
> nit: I'd probably change that line to reference a newer version as well, if we're removing the testing for it: https://github.com/aiven/myhoard/blob/master/README.md?plain=1#L791

uu good catch :+1:

packi (approved):

PR #149, opened 2023-03-07
Rebuild tables without primary key or with invalid dates

kmichel-aiven:
### About this change: What it does, why it matters

If the SQL mode or primary keys requirement was changed to something more strict since the tables were originally created, rebuilding the tables will fail.

Rebuilding essentially behaves as if the CREATE TABLE statement was re-executed.

To fix that, lower the requirements while rebuilding the tables.

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -450,6 +450,15 @@ class RestoreCoordinator(threading.Thread):
         self._ensure_mysql_server_is_started(with_binlog=False, with_gtids=False)
         # Rebuilding very large tables can be slow
         with self._mysql_cursor(timeout=3600.0) as cursor:
+            cursor.execute("SELECT @@SESSION.sql_mode")
+            unwanted_sql_modes = {"STRICT_ALL_TABLES", " STRICT_TRANS_TABLES", "NO_ZERO_DATE", "NO_ZERO_IN_DATE"}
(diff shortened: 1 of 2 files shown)


myhoard/restore_coordinator.py
```diff
@@ -450,6 +450,15 @@ def rebuild_tables(self) -> None:
         self._ensure_mysql_server_is_started(with_binlog=False, with_gtids=False)
         # Rebuilding very large tables can be slow
         with self._mysql_cursor(timeout=3600.0) as cursor:
+            cursor.execute("SELECT @@SESSION.sql_mode")
+            unwanted_sql_modes = {"STRICT_ALL_TABLES", " STRICT_TRANS_TABLES", "NO_ZERO_DATE", "NO_ZERO_IN_DATE"}
+            sql_mode = cursor.fetchone()["@@SESSION.sql_mode"]
+            lenient_sql_mode = ",".join([mode for mode in sql_mode.split(",") if mode not in unwanted_sql_modes])
+            if lenient_sql_mode != sql_mode:
```
packi, line 457: Don't we need some sort of sorting to ensure we're not tripped-up on different sort-orders?

kmichel-aiven: the `unwanted_sql_modes` is an unsorted set, but the split and comprehension list to edit the `sql_mode` itself keep the existing order, so I don't expect any issue there.

packi (approved):

PR #150, opened 2023-03-10
Disable innodb_strict_mode when rebuilding tables

kmichel-aiven:

If the option was already disabled in the session that created the table, then we need to also disable it when rebuilding the table or the rebuild can fail with:

`Row size too large. The maximum row size for the used table type, not counting BLOBs, is 8126.`


alanfranz (approved):
LGTM

PR #151, opened 2023-03-10
Extend rebuild tables timeout to 12h

kmichel-aiven:

The operation can easily take multiple hours on large tables.


alanfranz (approved):
LGTM.

PR #153, opened 2023-03-28
If there is a previous backup to base the time off we still need to s…

joshua-rule:
…et the backup hour

 so that the backup time can be changed

[PM-2995]

[PM-2995]: https://aiven.atlassian.net/browse/PM-2995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,15 +189,17 @@ than 24 hours this is used as base for calculating the backup times. E.g. if
 backup interval was 6 hours and backup hour was 4, backups would be taken at
 hours 4, 10, 16 and 22.
 
+**backup_minute**
+The minute of hour at which to take new full backup.
+
 **backup_interval_minutes**
 
 The interval in minutes at which to take new backups. Individual binary logs
 are backed up as soon as they're created so there's usually no need to have
 very frequent full backups.
-
-**backup_minute**
-
-The minute of hour at which to take new full backup.
+Note: If this value is not does not have a factor of 1440 (1 day) then the backup_hour and backup_minute
+settings cannot be changed once the first backup has been taken, as having a cycle not as a multiple of days
+means that the hour and minute of the backup will not be the same each day.
 
 **forced_binlog_rotation_interval**
 
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -872,15 +872,22 @@ class Controller(threading.Thread):
         backup_interval_minutes = self.backup_settings["backup_interval_minutes"]
         backup_hour = self.backup_settings["backup_hour"]
         backup_minute = self.backup_settings["backup_minute"]
+        day_in_minutes = 1440
 
         previous_normalized = self._previous_normalized_backup_timestamp()
-        # in case we have a previous backup time, we can use it as starting point
+
+        # If we have a previous backup we use this to base our current normalized datetime off
+        # this allows backup intervals of greater than a day
         if previous_normalized:
             normalized = datetime.datetime.fromisoformat(previous_normalized)
+            # If the interval is in days then we can change the time of day backups are taken
+            if backup_interval_minutes % day_in_minutes == 0:
+                normalized = normalized.replace(hour=backup_hour, minute=backup_minute)
         else:
-            if normalized.hour < backup_hour or (normalized.hour == backup_hour and normalized.minute < backup_minute):
-                normalized = normalized - datetime.timedelta(days=1)
-            normalized = normalized.replace(hour=backup_hour, minute=backup_minute, second=0, microsecond=0)
+            normalized = now.replace(hour=backup_hour, minute=backup_minute, second=0, microsecond=0)
+
+        if normalized > now:
+            normalized = normalized - datetime.timedelta(days=1)
 
         while normalized + datetime.timedelta(minutes=backup_interval_minutes) <= now:
             normalized = normalized + datetime.timedelta(minutes=backup_interval_minutes)
--- a/test/test_controller.py
+++ b/test/test_controller.py
@@ -1357,6 +1357,70 @@ def test_periodic_backup_based_on_exceeded_intervals(time_machine, master_contro
     while_asserts(lambda: streaming_binlogs(m_controller, 3), timeout=10)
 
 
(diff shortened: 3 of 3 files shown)


myhoard/controller.py
```diff
@@ -880,7 +880,8 @@ def _current_normalized_backup_timestamp(self) -> str:
         else:
             if normalized.hour < backup_hour or (normalized.hour == backup_hour and normalized.minute < backup_minute):
                 normalized = normalized - datetime.timedelta(days=1)
-            normalized = normalized.replace(hour=backup_hour, minute=backup_minute, second=0, microsecond=0)
+        
+        normalized = normalized.replace(hour=backup_hour, minute=backup_minute, second=0, microsecond=0)
```
aiven-amartin, line 884: I think you need to pull the whole if statement out of the else, as we need to do the date math to make sure our window isn't > 24h.

aiven-amartin: It also would be worth excavating if the actual bug is just that `_previous_normalized_backup_timestamp` parses all streams, rather than just active basebackup ones. I.e. I expect the reason we check the old one is so we don't change the request time while a stream is active, but then we get stuck in this perma-loop. I think we should just remove the state checking behaviour, just would need to validate how things behave when we change this in the middle of a backup (would not want to abort).

joshua-rule: It needs to check the previous backup time because if the backup interval is greater than 24 hours the previous logic will still run the backup at least once a day.

I started trying to solve the problem more generally but loads of problems arise because the interval is in minutes so the time drifts every day making the backup hour/minute number kind of meaningless.

I think maybe the simplest but not so nice way to fix our immediate problem with this is to check if the interval is greater than 24 hours and use the current logic to allow intervals greater than 24 hours, otherwise use the old logic so you can change the backup time.

Otherwise if we enforced some kind of allowed values on the interval, like hourly then if you get past 24 hours you can only have values that have 24 hours as a factor this all becomes a lot more sensible.

aiven-amartin: Makes sense to me. Can just update the documentation to say for intervals != 24h, the backup hour/minute configuration is not respected, as even if its less than 24hrs it doesn't really make sense either.

aiven-amartin: Ideally we would reject configuration that sets both sets to non-defaults but that would be a breaking change now I guess.

myhoard/controller.py
```diff
@@ -875,7 +875,7 @@ def _current_normalized_backup_timestamp(self) -> str:
 
         previous_normalized = self._previous_normalized_backup_timestamp()
         # in case we have a previous backup time, we can use it as starting point
-        if previous_normalized:
+        if previous_normalized and backup_interval_minutes > 1440:
```
aiven-amartin, line 878:
```suggestion
        if previous_normalized and backup_interval_minutes != 1440:
```

And this magic should probably be a magic value from somewhere, but that can be a follow up I guess. Also `>` is good enough for now I think.

README.md
```diff
@@ -189,15 +189,16 @@ than 24 hours this is used as base for calculating the backup times. E.g. if
...
 **backup_interval_minutes**
 
 The interval in minutes at which to take new backups. Individual binary logs
 are backed up as soon as they're created so there's usually no need to have
 very frequent full backups.
-
-**backup_minute**
-
-The minute of hour at which to take new full backup.
+Note: If this value is set to > 1440 (1 day) then the backup_hour and backup_minute
```
aiven-amartin, line 200:
```suggestion
Note: If this value is set to anything other than 1440 (1 day) then the backup_hour and backup_minute
```

README.md
```diff
@@ -189,15 +189,16 @@ than 24 hours this is used as base for calculating the backup times. E.g. if
...
+settings cannot be changed once the first backup has been taken
```
aiven-amartin, line 201:
```suggestion
settings cannot be changed once the first backup has been taken, as having a cycle != 1 day
means that the hour and minute of the backup will not be the same each day.
```

kmichel-aiven (changes requested):

README.md
```diff
@@ -189,15 +189,17 @@ than 24 hours this is used as base for calculating the backup times. E.g. if
...
 **backup_interval_minutes**
 
 The interval in minutes at which to take new backups. Individual binary logs
 are backed up as soon as they're created so there's usually no need to have
 very frequent full backups.
-
-**backup_minute**
-
-The minute of hour at which to take new full backup.
+Note: If this value is set to anything other than 1440 (1 day) then the backup_hour and backup_minute
```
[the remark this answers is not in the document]
joshua-rule: yea we can allow multiples now fairly easily, have made that change

joshua-rule:
looks like older versions of xtrabackup have disappeared from the repo, didn't get a chance to look into why that is.

test/test_controller.py
```diff
@@ -1357,6 +1357,71 @@ def flush_binlogs():
...
+
+    time_machine.move_to("2023-01-03T06:00:00+00:00")
+    while_asserts(lambda: streaming_binlogs(m_controller, 2), timeout=10)
+
+    # generate more data for third backup
+    flush_binlogs()
+
+    # After second backup, the next scheduled one should be at 2023-02-04 03:00:00, but let's change
+    # backup_interval_minutes to 2880 (48 hours)
+    # m_controller.backup_settings["backup_interval_minutes"] = 2880
```
kmichel-aiven, line 1406: I don't think this line should be commented ?

Given the comment, it should be there, and if it shouldn't, then it should be fully removed.

kmichel-aiven (changes requested):

kmichel-aiven:
> looks like older versions of xtrabackup have disappeared from the repo, didn't get a chance to look into why that is.

That's not ideal :(

Well, since this PR is purely internal logic, as long as the 8.0.30 tests pass, I'm fine with merging it, we can solve the repo problem in a separate PR.

kmichel-aiven (approved):

PR #154, opened 2023-03-31
Enforce using secure db connections

ettanany:
Ensure various backup management tasks can be done when the server requires encrypted connections.


alexole:
I don't think we should enforce SSL for `myhoard`, this is an open source project and I assume there will be valid scenarios where SSL is not needed. It should be possible to switch on SSL, but not mandatory if one ops out.

PR #155, opened 2023-04-04
Prevent myhoard starting more than one backup stream

nicois:
Multiple backup streams add complexity and instability, increasing load on the server for negligible benefit. It is better to complete the current backup before devoting resources to beginning a second stream.

This also has the side-effect of better handling situations where one stream appears to fail, then recovers, and interferes with a new backup stream: until the first one is finished, a second one will not start, after which the first one is gone.

A "terminated" backup stream (ie: one which is completed, closed or broken) does not count towards this limit; only one which is not yet in one of these terminal states is.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -252,6 +252,10 @@ class BackupStream(threading.Thread):
         yield
         self.stop()
 
+    def is_in_terminal_state(self) -> bool:
+        """Returns if the stream has finished operating, successfully or not"""
+        return bool(self.state.get("broken_info") or self.state.get("closed_info") or self.state.get("completed_info"))
+
     def activate(self) -> None:
         with self.lock:
             if self.mode != self.Mode.promoted:
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -802,11 +802,12 @@ class Controller(threading.Thread):
             self.state_manager.update_state(promote_details=promote_details)
 
     def _create_new_backup_stream_if_requested_and_max_streams_not_exceeded(self):
-        # Only ever have two open backup streams. Uploading binlogs to more streams than that is
-        # unlikely to improve the system behavior. We'll create new backup stream once the latter
-        # one catches up with the first, the first is marked as closed, and removed from our list.
-        if len(self.backup_streams) >= 2:
-            return
+        # Only ever have one "active" backup stream.
+        # Having multiple backup streams adds complexity, increases node resource usage, and
+        # more easily triggers various latent bugs.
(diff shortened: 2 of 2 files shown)


rikonen (commented):
I'm not sure what the intended behavior with the streams with this change is exactly as the streams are marked as completed / closed by the controller when we have multiple concurrent streams and certain conditions are met and it looks to me like this would just not start multiple streams at all, i.e. would not create new backups. But also worth keeping in mind that in order for point-in-time recovery to work we must have the previous stream in active state and uploading binlogs until the new one is fully complete and we must not break PITR so I'm not convinced we can make any changes related to this.

nicois:
@rikonen , when I create a PITR fork of a service in a test environment, I do not see a second stream being created on the primary service. The only activity appears to be a forced binlog flush. 
Are there specific situations where a second backup stream is required, or could it be that these days a single stream is sufficient?
```
May 04 00:30:33 nick-farrell-test-n1017 myhoard[3169]: 2023-05-04 00:30:33,358        MainThread        Controller        INFO        Rotating binlog due to external request
May 04 00:30:33 nick-farrell-test-n1017 myhoard[3169]: 2023-05-04 00:30:33,365        MainThread        WebServer        INFO        Waiting up to 3.0 seconds for upload of 5 to complete
May 04 00:30:33 nick-farrell-test-n1017 myhoard[3169]: 2023-05-04 00:30:33,366        Thread-1        BinlogScanner        INFO        New binlog '/srv/aiven-persistent/1017-service/mysqlhome/8/binlogs/binlog.000005' (241 bytes) in 0.00 seconds, found 0 GTID ranges: []
May 04 00:30:33 nick-farrell-test-n1017 myhoard[3169]: 2023-05-04 00:30:33,824        Thread-3        BackupStream/2023-05-04T00:20:30.782913+00:00Z_bf5060b2-94bc-4693-b5b9-22e2efad652f        INFO        Successfully uploaded local binlog index 5 as remote index 4 in 0.5 seconds
May 04 00:30:33 nick-farrell-test-n1017 myhoard[3169]: 2023-05-04 00:30:33,867        MainThread        WebServer        INFO        Log 5 was backed up in 0.5 seconds
```

alexole:
I don't think this achieves the stated goal. Streams get `mark_as_completed()` called almost immediately when they're the only one (they keep uploading binlogs, but completed_info is set), so `is_in_terminal_state()` returns True and a second stream starts anyway. Net effect is the same concurrent overlap as before, just gated on a different state.
The concurrent streams exist for a reason: PITR relies on them to keep binlog coverage continuous. Any stricter gating would first need to guarantee there's no binlog gap when handing off between streams.

PR #156, opened 2023-04-04
Support using secure db connections [BF-1965]

ettanany:
Ensure various backup management tasks can be done when the server requires encrypted connections.

### DIFF
--- a/myhoard.json
+++ b/myhoard.json
@@ -37,9 +37,10 @@
         "binlog_prefix": "/var/lib/mysql/binlog",
         "client_params": {
             "host": "127.0.0.1",
-            "password": "NgLqvU8gbWCtfJWJPy",
+            "password": "f@keP@ssw0rd",
             "port": 3306,
-            "user": "root"
+            "user": "root",
+            "require_ssl": false
         },
         "config_file_name": "/etc/my.cnf",
         "data_directory": "/var/lib/mysql",
--- a/myhoard/util.py
+++ b/myhoard/util.py
@@ -366,10 +366,10 @@ def are_gtids_in_executed_set(gtid_executed, ranges, *, exclude_uuid=None):
 
 
 @contextlib.contextmanager
-def mysql_connection(*, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user):
-    ssl = None
-    if ca_file:
-        ssl = {"ca": ca_file}
+def mysql_connection(
+    *, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user, require_ssl=False
+):
+    ssl = {"ca": ca_file} if ca_file else {"require": True} if require_ssl else None
     connection = pymysql.connect(
         charset="utf8mb4",
         connect_timeout=timeout,
@@ -390,9 +390,11 @@ def mysql_connection(*, ca_file=None, db="mysql", host="127.0.0.1", password, po
 
 
 @contextlib.contextmanager
-def mysql_cursor(*, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user):
+def mysql_cursor(
+    *, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user, require_ssl=False
+):
     with mysql_connection(
-        ca_file=ca_file, db=db, host=host, password=password, port=port, timeout=timeout, user=user
+        ca_file=ca_file, db=db, host=host, password=password, port=port, timeout=timeout, user=user, require_ssl=require_ssl
     ) as connection:
         with connection.cursor() as cursor:
             yield cursor
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -327,7 +327,13 @@ def fixture_myhoard_config(default_backup_site, mysql_master, session_tmpdir):
         "http_port": get_random_port(start=3000, end=30000),
         "mysql": {
             "binlog_prefix": mysql_master.config_options.binlog_file_prefix,
-            "client_params": {"host": "127.0.0.1", "password": "NgLqvU8gbWCtfJWJPy", "port": 3306, "user": "root"},
+            "client_params": {
+                "host": "127.0.0.1",
+                "password": "f@keP@ssw0rd",
+                "port": 3306,
+                "require_ssl": False,
+                "user": "root",
+            },
             "config_file_name": mysql_master.config_name,
             "data_directory": mysql_master.config_options.datadir,
             "relay_log_index_file": mysql_master.config_options.relay_log_index_file,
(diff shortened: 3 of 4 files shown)


test/conftest.py
```diff
@@ -327,7 +327,13 @@ def fixture_myhoard_config(default_backup_site, mysql_master, session_tmpdir):
         "http_port": get_random_port(start=3000, end=30000),
         "mysql": {
             "binlog_prefix": mysql_master.config_options.binlog_file_prefix,
-            "client_params": {"host": "127.0.0.1", "password": "NgLqvU8gbWCtfJWJPy", "port": 3306, "user": "root"},
+            "client_params": {
+                "host": "127.0.0.1",
+                "password": "NgLqvU8gbWCtfJWJPy",
```
Mulugruntz, line 332: It was already in the source, but I'd add a comment to precise that it's fake.
If it's not it must be changed + rotated.

test/test_util.py
```diff
@@ -3,19 +3,72 @@
...
+            "password": "QR2$9Up2FAqF",
+            "port": 3306,
+            "require_ssl": True,
+            "user": "root",
+        },
+        {
+            "ca_file": "ca-bundle.crt",
+            "db": "data",
+            "host": "localhost",
+            "password": "QR2$9Up2FAqF",
```
Mulugruntz, line 39: same for all passwords.

test/test_util.py
```diff
@@ -3,19 +3,72 @@
...
+import pymysql
 import pytest
 import random
 import subprocess
 
 pytestmark = [pytest.mark.unittest, pytest.mark.all]
 
 
+@patch("pymysql.connect")
+@pytest.mark.parametrize(
```
Mulugruntz, line 21: we don't have the case where both ca_file and require_ssl are provided.

test/test_util.py
```diff
@@ -3,19 +3,72 @@
...
+            "timeout": 10.0,
+            "user": "admin",
+        },
+    ],
+)
+def test_mysql_cursor(connect_mock, conn_options):
+    with myhoard_util.mysql_cursor(**conn_options):
+        pass
+
+    timeout = conn_options.get("timeout", 4.0)
```
Mulugruntz, line 50: Where does this `4.0` come from? It should probably be a constant in the code. If it's a magic const, that should be improved. Then you can import it in the test and use it here.

myhoard/util.py
```diff
@@ -366,8 +366,12 @@ def are_gtids_in_executed_set(gtid_executed, ranges, *, exclude_uuid=None):
...
 @contextlib.contextmanager
-def mysql_connection(*, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user):
+def mysql_connection(
+    *, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user, require_ssl
+):
     ssl = None
+    if require_ssl:
+        ssl = {"require": True}
     if ca_file:
         ssl = {"ca": ca_file}
```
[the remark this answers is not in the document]
ettanany: `ssl = {"ca": ca_file}` would enforce SSL and cause the certificate to be verified, while `ssl = {"require": True}` would only enforce SSL.

test/test_util.py
ettanany: I just used the same default from `util.py` (`DEFAULT_MYSQL_TIMEOUT = 4.0`).

test/conftest.py
ettanany: It's fake (same one that is used in `myhoard.json` file).

myhoard.json
```diff
@@ -37,9 +37,10 @@
         "binlog_prefix": "/var/lib/mysql/binlog",
         "client_params": {
             "host": "127.0.0.1",
-            "password": "NgLqvU8gbWCtfJWJPy",
+            "password": "f@keP@ssw0rd",
```
Mulugruntz, line 40: 👍

test/test_util.py
ettanany: I personally do not like sharing constants between actual code and tests (for obvious reasons of making tests reliable, i.e. if we expect a timeout to be 10s, then our tests should enforce that, which means my tests should catch (fail because of) any change of my actual code that changes it to something else other than 10s). With that said, my approach is to either hard code something like this in my tests if it's only used once or extract it to tests-specific constants in case we need to reuse it.

myhoard/util.py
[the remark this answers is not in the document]
ettanany: From my testing with this library, there is only one way to enable SSL, which assigning anything that is not `None` to `ssl`. I will add a comment to clarify this.

test/test_util.py
Mulugruntz: Okay, your explanation also makes sense and I'll resolve this thread! Thanks!

test/test_util.py
Mulugruntz: I think we should have the case, to show that the behaviour works as intended, even when the input parameters are a bit overzealous. And to show that the ca_file takes precedence.

myhoard/util.py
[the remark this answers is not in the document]
Mulugruntz: > unless I missed something, `{"require": False}` also enabled SSL when I tested it

😱

test/test_util.py
ettanany: :heavy_check_mark:

Mulugruntz (approved):
Let's go! At least, for me.

myhoard/util.py
```diff
@@ -366,10 +366,10 @@ def are_gtids_in_executed_set(gtid_executed, ranges, *, exclude_uuid=None):
...
 
 @contextlib.contextmanager
-def mysql_connection(*, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user):
-    ssl = None
-    if ca_file:
-        ssl = {"ca": ca_file}
+def mysql_connection(
+    *, ca_file=None, db="mysql", host="127.0.0.1", password, port, timeout=DEFAULT_MYSQL_TIMEOUT, user, require_ssl=False
+):
+    ssl = {"ca": ca_file} if ca_file else {"require": True} if require_ssl else None
```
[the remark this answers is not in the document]
alexole: If for any reason it does not enable SSL there is another possibility to use `ssl_ca` parameter for the connection and it will enable SSL (check the code I've posted)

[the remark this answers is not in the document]
ettanany: We need to pass it a truthy value because of this line: https://github.com/PyMySQL/PyMySQL/blob/885841f3fee416c222a75d83a81f74d3dcd71b51/pymysql/connections.py#L279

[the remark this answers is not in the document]
ettanany: > If for any reason it does not enable SSL there is another possibility to use `ssl_ca` parameter for the connection and it will enable SSL (check the code by the link I've posted)

My understanding is that we want to support enabling SSL without passing `ca_file`. Or I am missing something?

alexole (approved):

PR #159, opened 2023-05-08
Use inplace algorithm when rebuilding tables to save space [BF-1991]

alexole:


alexole:
This should already be used implicitly if possible. Thus this change is not needed.

PR #160, opened 2023-05-29
Backup is not broken on disk full [BF-2051]

Mulugruntz:
Mainly three things in this PR:

- When restoring a backup fails because of Disk Full, the backup is no longer marked as broken.
- Developer experience on M1/arm: `make build-setup-specific-image` now works on more architectures than `x86_64`.
- Python 3.10 was missing in the CI. We had 3.9 and 3.11, whereas the linters were running on 3.10.

[BF-2051]

[BF-2051]: https://aiven.atlassian.net/browse/BF-2051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


myhoard/basebackup_restore_operation.py
```diff
@@ -313,6 +322,10 @@ def _process_xbstream_output_line(self, line, _stream_name):
         if not line:
             return
 
+        if self._process_xtrabackup_output_line_no_space_left(line):
+            self.log.error("xbstream: %r", line)
+            raise DiskFullError("No space left on device. Cannot complete xbstream-extract!")
```
rikonen, line 327: These blocks are repetitive. The `_process_xtrabackup_output_line_no_space_left` function already logs the line so logging it here is not necessary. Also error could be raised from the function so that the callers don't need to do anything extra. (In practice only the `xbstream` phase is in practice potentially running out of disk so doing this for others is not that relevant but might be a good idea just in case.)

myhoard/controller.py
```diff
@@ -1174,7 +1174,9 @@ def _handle_mode_restore(self):
         self._purge_old_binlogs(mysql_maybe_not_running=True)
         self._process_local_binlog_updates()
         self._extend_binlog_stream_list()
-        if self.restore_coordinator.phase == RestoreCoordinator.Phase.failed_basebackup:
+        if self.restore_coordinator.phase is RestoreCoordinator.Phase.failed_disk_full:
+            self._switch_basebackup_if_possible()
```
rikonen, line 1178: Trying another backup isn't probably very useful. Since we don't want the controller to do anything special in this case we could omit the new restore coordinator state and set the state in `RestoreCoordinator` to `RestoreCoordinator.Phase.failed` when encountering `DiskFullError`

myhoard/errors.py
```diff
@@ -19,3 +19,7 @@ def __init__(self, backup_site_name, known_backup_sites):
 
     def __str__(self):
         return f"backup site {self.backup_site_name} unknown (sites: {self.known_backup_sites!r})"
+
+
+class DiskFullError(Exception):
+    """Raised when disk is full.""" ""
```
rikonen, line 25: Extra ` ""` at the end

rikonen (dismissed):

myhoard/controller.py
Mulugruntz: The question was more... "maybe the last one is too big, but an older one is small enough to be restored to?"

But I'm okay with just failing and not trying other basebackups.

WDYT?

myhoard/basebackup_restore_operation.py
Mulugruntz: I think `_process_xtrabackup_output_line_no_space_left` should not log. It's a predicate function, it should not have logging side-effects. The reason why it logs, is because I reused the existing patterns.

The repetitive blocks then "become" correct, as `_process_xtrabackup_output_line_no_space_left` should not have more responsibilities.
It would also be possible to have yet another method, which would be something like:
```python
    def _raise_if_disk_full(self, line: str, bin_name: str, caller_id: str) -> None:
        if self._process_xtrabackup_output_line_no_space_left(line):
            self.log.error("%s: %r", bin_name, line)
            raise DiskFullError("No space left on device. Cannot complete %s!", caller_id)
```

But it really looks like a "util" function, to force DRY-ness. I'm not super excited about that, and would prefer having these 3 blocks. Especially as they could have different logic and should not depend on the other binary's stdout/stderr.

myhoard/controller.py
rikonen: If we restore an older backup all the binlogs from both the old and new backup are still applied and it ends up with the same data. Maybe there is some theoretical case where old basebackup is smaller and the binlogs won't make it as big as the more recent basebackup itself but I don't really see how that would happen.

myhoard/basebackup_restore_operation.py
rikonen: I would go with the extra function to avoid duplication but I can anyway leave this pr for the team actually working on this component so no need to mind me. I just had a quick glance as I was curious to see if we can easily detect the "out of disk space" situation (seems like we can).

packi: I tend to agree with @rikonen here, a separate function makes sense as the log-message is basically the same.

packi (dismissed):
I've re-opened the comment from Raouli, I think we should extract it into a function. Not as a requested change (NIT), but please keep the actual change in a separate commit. If we're mixing typing with other changes, reverting an isolated fix becomes a pain and merging it becomes more risky.

myhoard/basebackup_restore_operation.py
Mulugruntz: Changed in new version. 🚀

test/conftest.py
```diff
@@ -298,6 +301,64 @@ def fixture_standby2_controller(session_tmpdir, mysql_standby2, default_backup_s
...
+        path: The path to mount the tmpfs filesystem at (will create a subdirectory there).
+        size: The size of the tmpfs filesystem in megabytes.
+
+    Yields:
+        The path the tmpfs filesystem was mounted at.
+    """
+    sub_dir = path / random_basic_string(20, prefix="small_disk_")
+    try:
+        sub_dir.mkdir(parents=True, exist_ok=True)
+        subprocess.check_call(["sudo", "mount", "-t", "tmpfs", "-o", f"size={size}m", "tmpfs", str(sub_dir)])
```
packi, line 354: Tests requiring `sudo` are probably not the best idea, but we can leave it in for now as it runs on github. But I would probably split the usage of this into a separate integration test to make sure we can still run the bulk of the tests locally without any special treatment. This way we could also split out the fixtures to this special file, as they are used only from one file.

test/conftest.py
```diff
@@ -298,6 +301,64 @@ def fixture_standby2_controller(session_tmpdir, mysql_standby2, default_backup_s
...
+) -> Iterator[tuple[Controller, MySQLConfig]]:
+    sub_dir = Path(session_tmpdir().strpath) / "small_disk"
+    with mount_tmpfs(path=sub_dir, size=8) as tmpfs_dir:
+        controller = build_controller(
+            Controller,
+            default_backup_site=default_backup_site,
+            mysql_config=mysql_empty,
+            session_tmpdir=lambda: LocalPath(tmpfs_dir),
+        )
+        try:
```
packi, line 334: I don't think try..finally work as intended in fixtures, there has been a discussion about that in a recent Friday demo.

test/test_controller.py
```diff
@@ -1722,6 +1728,254 @@ def test_check_if_get_backup_list_fills_attributes_based_on_missing_files(
...
+    """Test a backup and restore that fails restoring because the disk is full."""
+    empty_controller_in_small_disk[1].connect_options["password"] = master_controller[1].connect_options["password"]
+    _populate_table(master_controller[1], "test")
+
+    backup_streams = do_backup(controller=master_controller[0])
+    do_restore(
+        target_controller=empty_controller_in_small_disk[0],
+        backup_streams=backup_streams,
+        caplog=caplog,
+        fail_because_disk_full=True,
```
packi, line 1772: This only checks for the log, not the actual flag of the backup.

packi (dismissed):
Thanks for splitting the commits. A couple of other things I noticed while giving it a closer look (see comments).

myhoard/restore_coordinator.py
```diff
@@ -419,7 +419,12 @@ def restore_basebackup(self) -> None:
             free_memory_percentage=self.free_memory_percentage,
         )
         try:
-            self.basebackup_restore_operation.restore_backup()
+            try:
+                self.basebackup_restore_operation.restore_backup()
+            except DiskFullError:
+                self.stats.increase("myhoard.disk_full_errors")
+                self.update_state(phase=self.Phase.failed)
+                raise
```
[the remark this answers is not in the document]
Mulugruntz: No, this is only if we try several times and reached the max try. And because we'll make it fail from the first time, it won't happen.

test/conftest.py
Mulugruntz: Agreed. It's @rikonen 's suggestion, which works great for what we need. It's definitely not ideal but shouldn't be a blocker for now :-).

myhoard/restore_coordinator.py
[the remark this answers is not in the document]
packi: So, I had another look at it. If you re-raise here, the exception block below gets called and if it fails enough times, it will still mark it as broken because `self.update_state(phase=self.Phase.failed_basebackup)` will get called and this will mark it as broken in the controller. So, at least to me, it seems like you need to re-catch the exception in the outer `try: .. catch:` block as well.

test/test_controller.py
```diff
@@ -1722,6 +1738,259 @@ def test_check_if_get_backup_list_fills_attributes_based_on_missing_files(
...
+
+        target_controller.restore_backup(site=bs.site, stream_id=bs.stream_id)
+
+        if fail_because_disk_full:
+            flow_tester.wait_for_disk_full_being_logged()
+
+            # Check that we have backups, but none of them are broken.
+            current_backups = sort_completed_backups(target_controller.state["backups"])
+            assert current_backups
+            assert all(b["broken_at"] is None for b in current_backups)
```
packi, line 1921: You should probably also check the state of the `restore_coordinator`, it should be `failed` and not `failed_basebackup` (which I assume it is if you run this check `MAX_BASEBACKUP_ERRORS` times).

packi (dismissed):
See the two comments. I'd still split the new tests that require sudo into their own file, including the fixtures.

myhoard/restore_coordinator.py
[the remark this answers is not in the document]
Mulugruntz: Hmmm...

The first time it'll fail, it'll mark the state as `Phase.failed`.

Then, in `restore_coordinator.py:349`, we'll have:
```python
                if self.phase in {self.Phase.completed, self.Phase.failed, self.Phase.failed_basebackup}:
                    break
```

which will end the running loop of the restore coordinator.

So, the implementation should be correct as it is now.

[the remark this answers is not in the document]
packi: OK, I also see that we're not re-starting the restore_coordinator in any of the cases so we should be good. Thanks for double-checking!

packi (approved):

PR #162, opened 2023-06-27
Disable foreign keys checks when rebuilding tables [BF-2138]

alexole:
It might happen that rebuilding will fail with foreign key constraint error, so we should disable it while rebuilding.

[BF-2138]

[BF-2138]: https://aiven.atlassian.net/browse/BF-2138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Mulugruntz (approved):
Looks good as it is.
But at the end of the rebuild, we could have missing FKs? Or am I missing something?

Edit: @packi gave me more context. No need for explanation anymore. :-)

packi (approved):

PR #167, opened 2023-08-09
Fix flakiness in test_backup_and_restore_fail_on_disk_full

kathia-barahona:
`test/local/test_controller.py::test_backup_and_restore_fail_on_disk_full` was failing due to a race condition. Now it waits till the coordinator reaches `failed` phase and afterwards checks for disk full log.

[BF-2161]

### DIFF
--- a/test/helpers/filesystem.py
+++ b/test/helpers/filesystem.py
@@ -1,7 +1,6 @@
 # Copyright (c) 2023 Aiven, Helsinki, Finland. https://aiven.io/
-from contextlib import contextmanager, suppress
+from contextlib import contextmanager
 from pathlib import Path
-from py.path import local as LocalPath
 from test import random_basic_string
 from typing import Iterator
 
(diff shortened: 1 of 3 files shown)


test/helpers/filesystem.py
```diff
@@ -26,7 +25,4 @@ def mount_tmpfs(path: Path, *, megabytes: int) -> Iterator[Path]:
 
         yield sub_dir
     finally:
-        # Delete all files in the tmpfs filesystem before unmounting it.
-        with suppress(Exception):
-            LocalPath(sub_dir).remove(rec=1)
-            subprocess.check_call(["sudo", "umount", str(sub_dir)])
```
kathia-barahona, line 32: ended up removing this, because it was raising an exception and pytest was  complaining since it was not able to delete temporary directories

packi (approved):
LGTM, thanks for fixing this!

PR #171, opened 2023-09-11
support backup preservation  [BF-2219]

kathia-barahona:

Supports request for preserving/holding a backup till a particular date, this way we can guarantee that the backup is not deleted during a restoration request.

If the oldest backup is being preserved, myhoard will not purge any backup but will keep generating new backups. Purging is resumed till the preservation date is met.

## How to request a backup preservation?

For preserving a backup, a request to the HTTP API needs to be performed. For example:
```
PUT /backup/valid_stream_id_1234/preserve
{
    "preserve_until": "2023-09-12T13:30:00"
}
```

This will trigger the controller on fetching the respective backup stream and storing `preserved_info.json` on the object storage, it will contain the request  `preserve_until`. After storing the file, all backups metadata are immediately refreshed and myhoard considers updated `preserve_until` values before trying to purge old backups.

## How to cancel a backup preservation?
```
PUT /backup/valid_stream_id_1234/preserve
{
    "preserve_until": null,
}
```
This will trigger the controller on fetching the respective backup stream and storing `preserved_info.json` on the object storage, but will store an empty "preseve_until" value.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -140,6 +140,7 @@ class BackupStream(threading.Thread):
         normalized_backup_time: Optional[str]
         pending_binlogs: List[BinlogInfo]
         prepare_details: Dict
+        preserved_info: Dict
         # Set of GTIDs that have been stored persistently to file storage.
         remote_gtid_executed: GtidExecuted
         remote_read_errors: int
@@ -238,6 +239,7 @@ class BackupStream(threading.Thread):
             "normalized_backup_time": normalized_backup_time,
             "pending_binlogs": list(binlogs) if binlogs else [],
             "prepare_details": {},
+            "preserved_info": {},
             # Set of GTIDs that have been stored persistently to file storage.
             "remote_gtid_executed": {},
             "remote_read_errors": 0,
@@ -442,6 +444,33 @@ class BackupStream(threading.Thread):
             },
         )
         self.wakeup_event.set()
+
+    def mark_preservation(self, preserve_until: Optional[datetime]) -> None:
+        if preserve_until:
+            self.log.info("Marking stream %s preservation to %s.", self.stream_id, preserve_until)
+        else:
+            self.log.info("Removing preservation for stream %s.", self.stream_id)
+
+        preserved_info = {
+            "preserve_until": preserve_until.isoformat() if preserve_until else None,
+            "server_id": self.server_id,
+        }
+
+        try:
+            key = self._build_full_name("preserved.json")
+            data = json.dumps(preserved_info)
+            metadata = make_fs_metadata(preserved_info)
+
+            file_storage = self.file_storage_setup_fn()
+            file_storage.store_file_from_memory(key, data.encode("utf-8"), metadata=metadata)
+            self.state_manager.update_state(preserved_info=preserved_info)
+        except Exception as ex:  # pylint: disable=broad-except
+            self.log.exception("Failed to update basebackup preservation")
+            self.stats.unexpected_exception(ex=ex, where="BackupStream._update_basebackup_preservation")
+            self.state_manager.update_state(remote_write_errors=self.state["remote_write_errors"] + 1)
+            self.stats.increase("myhoard.remote_write_errors")
+            # raise exception since update on preservation did not succeed
+            raise
 
     def _handle_pending_mark_as_closed(self) -> None:
         assert self.file_storage is not None
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -55,6 +55,7 @@ class Backup(TypedDict):
     closed_at: Optional[float]
     completed_at: Optional[float]
     broken_at: Optional[float]
+    preserve_until: Optional[str]
     recovery_site: bool
     stream_id: str
     resumable: bool
@@ -127,10 +128,12 @@ class Controller(threading.Thread):
         last_could_have_purged: float
         mode: "Controller.Mode"
         owned_stream_ids: List[int]
+        pending_preservation_requests: Dict[str, Optional[str]]
         promote_details: Dict[str, Any]
         promote_on_restore_completion: bool
         replication_state: Dict[str, GtidExecuted]
         restore_options: dict
+        stream_to_be_purged: Optional[str]
         server_uuid: Optional[str]
         uploaded_binlogs: list
 
@@ -210,10 +213,12 @@ class Controller(threading.Thread):
             "last_could_have_purged": time.time(),
             "mode": self.Mode.idle,
             "owned_stream_ids": [],
+            "pending_preservation_requests": {},
             "promote_details": {},
             "promote_on_restore_completion": False,
             "replication_state": {},
             "restore_options": {},
+            "stream_to_be_purged": None,
             "server_uuid": None,
             "uploaded_binlogs": [],
         }
@@ -264,6 +269,16 @@ class Controller(threading.Thread):
                 ):
                     return
             self.state_manager.update_state(backup_request=new_request)
+
+    def mark_backup_preservation(self, stream_id: str, preserve_until: Optional[datetime.datetime]) -> None:
+        backup_to_preserve = self.get_backup_by_stream_id(stream_id)
+        if not backup_to_preserve:
+            raise Exception(f"Stream {stream_id} was not found in completed backups.")
+
+        with self.lock:
+            current_requests = dict(self.state["pending_preservation_requests"])
+            current_requests[stream_id] = preserve_until.isoformat() if preserve_until else None
+            self.state_manager.update_state(pending_preservation_requests=current_requests)
 
     @property
     def mode(self) -> Mode:
@@ -567,6 +582,7 @@ class Controller(threading.Thread):
                 broken_info = {}
                 closed_info = {}
                 completed_info = {}
+                preserved_info = {}
                 for info in file_storage.list_iter(site_and_stream_id):
                     file_name = info["name"].rsplit("/", 1)[-1]
                     if file_name == "basebackup.xbstream":
@@ -585,18 +601,23 @@ class Controller(threading.Thread):
                         closed_info = parse_fs_metadata(info["metadata"])
                     elif file_name == "completed.json":
                         completed_info = parse_fs_metadata(info["metadata"])
+                    elif file_name == "preserved.json":
+                        preserved_info = parse_fs_metadata(info["metadata"])
 
                 if basebackup_info and basebackup_compressed_size:
                     basebackup_info = dict(basebackup_info, compressed_size=basebackup_compressed_size)
                 resumable = basebackup_info and basebackup_compressed_size
                 completed = resumable and completed_info
                 closed = completed and closed_info
+
+                preserve_until = preserved_info.get("preserve_until")
                 backups.append(
                     {
                         "basebackup_info": basebackup_info,
                         "broken_at": broken_info.get("broken_at"),
                         "closed_at": closed_info["closed_at"] if closed else None,
                         "completed_at": completed_info["completed_at"] if completed else None,
+                        "preserve_until": preserve_until,
                         "recovery_site": site_config.get("recovery_only", False),
                         "stream_id": site_and_stream_id.rsplit("/", 1)[-1],
                         "resumable": bool(resumable),
@@ -1086,11 +1107,18 @@ class Controller(threading.Thread):
             raise Exception("Defined upload site not present in list of non-recovery backup sites")
         return site_id, non_recovery_sites[site_id]
 
-    def _get_site_for_stream_id(self, stream_id):
+    def _get_site_for_stream_id(self, stream_id: str):
+        backup = self.get_backup_by_stream_id(stream_id)
+        if not backup:
+            KeyError(f"Stream {stream_id} not found in backups")
+        return backup["site"]
+
+    def get_backup_by_stream_id(self, stream_id: str):
         for backup in self.state["backups"]:
             if backup["stream_id"] == stream_id:
-                return backup["site"]
-        raise KeyError(f"Stream {stream_id} not found in backups")
+                return backup
+
+        return None
 
     def _get_restore_coordinator_state_file_and_remove_old(self):
         state_file_name = os.path.join(self.state_dir, "restore_coordinator_state.json")
@@ -1111,6 +1139,7 @@ class Controller(threading.Thread):
     def _handle_mode_active(self):
         self._cache_server_uuid_if_missing()
         self._set_uploaded_binlog_references()
+        self._handle_pending_preservation_requests()
         self._refresh_backups_list_and_streams()
         self._mark_periodic_backup_requested_if_interval_exceeded()
         self._create_new_backup_stream_if_requested_and_max_streams_not_exceeded()
@@ -1188,16 +1217,13 @@ class Controller(threading.Thread):
             )
 
     def _mark_failed_restore_backup_as_broken(self) -> None:
-        broken_backup = None
         failed_stream_id = self.state["restore_options"]["stream_id"]
-        backups = self.state["backups"]
-        for backup in backups:
-            if backup["stream_id"] == failed_stream_id:
-                broken_backup = backup
-                break
+        broken_backup = self.get_backup_by_stream_id(stream_id=failed_stream_id)
 
         if not broken_backup:
-            raise Exception(f"Stream {failed_stream_id} to be marked as broken not found in completed backups: {backups}")
+            raise Exception(
+                f'Stream {failed_stream_id} to be marked as broken not found in completed backups: {self.state["backups"]}'
+            )
 
         self._build_backup_stream(broken_backup).mark_as_broken()
 
@@ -1290,6 +1316,11 @@ class Controller(threading.Thread):
         if not backup["closed_at"]:
             return
 
+        # do not purge backup if its preserved
+        preserve_until = backup["preserve_until"]
+        if preserve_until and datetime.datetime.now(datetime.timezone.utc) < datetime.datetime.fromisoformat(preserve_until):
+            return
+
         if time.time() > backup["closed_at"] + self.backup_settings["backup_age_days_max"] * 24 * 60 * 60:
             self.log.info("Backup %r is older than max backup age, dropping it", backup["stream_id"])
         elif non_broken_backups_count > self.backup_settings["backup_count_max"]:
@@ -1304,8 +1335,19 @@ class Controller(threading.Thread):
             self.log.warning("Backup %r to drop is one of active streams, not dropping", backup["stream_id"])
             return
 
-        self._build_backup_stream(backup).remove()
         with self.lock:
+            self.state_manager.update_state(stream_to_be_purged=backup["stream_id"])
+
+        self._build_backup_stream(backup).remove()
+        # lock the controller, this way other requests do not access backups till backup is purged
+        with self.lock:
+            self.state_manager.update_state(stream_to_be_purged=None)
+            current_backups = [
+                current_backup
+                for current_backup in self.state["backups"]
+                if current_backup["stream_id"] != backup["stream_id"]
+            ]
+            self.state_manager.update_state(backups=current_backups)
             owned_stream_ids = [sid for sid in self.state["owned_stream_ids"] if sid != backup["stream_id"]]
             self.state_manager.update_state(owned_stream_ids=owned_stream_ids)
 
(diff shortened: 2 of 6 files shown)


Mulugruntz (dismissed):
Looks good and functional. The tests are also great!
However, there are 2 main points that might need refinement:
- the http endpoints could be better
- the named tuple `PreservationRequest` brings no value and obfuscates the code a bit.

On less important points (not blocking blocking):
- the wasteful iteration over `state["backups"]` to find the one we want with `stream_id` looks like it's not the right data structure. I'm not sure if there's a reason for it being a list.
- type hints missing here and there
- some boolean expressions

myhoard/backup_stream.py
```diff
@@ -437,6 +439,30 @@ def mark_as_closed(self) -> None:
...
+
+            file_storage = self.file_storage_setup_fn()
+            file_storage.store_file_from_memory(key, data.encode("utf-8"), metadata=metadata)
+        except Exception as ex:  # pylint: disable=broad-except
+            self.log.exception("Failed to update basebackup preservation")
+            self.stats.unexpected_exception(ex=ex, where="BackupStream._update_basebackup_preservation")
+            self.state_manager.update_state(remote_write_errors=self.state["remote_write_errors"] + 1)
+            self.stats.increase("myhoard.remote_write_errors")
+            # raise exception since update on preservation did not succeed
+            raise ex
```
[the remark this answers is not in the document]
kathia-barahona: The point here is returning the actual exception message to the API

myhoard/controller.py
```diff
@@ -265,6 +267,15 @@ def mark_backup_requested(
                     return
             self.state_manager.update_state(backup_request=new_request)
 
+    def mark_backup_preservation(self, preservation_request: PreservationRequest) -> None:
+        backup_to_preserve = self._get_backup_by_stream_id(preservation_request.stream_id)
+        if not backup_to_preserve:
+            raise Exception(f"Stream {preservation_request.stream_id} was not found in completed backups.")
```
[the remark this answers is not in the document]
kathia-barahona: Why? This is the pattern used in the entire code base.

myhoard/backup_stream.py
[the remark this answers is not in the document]
Mulugruntz: If we just use `raise`, the initial message and backtrace will be preserved.

myhoard/controller.py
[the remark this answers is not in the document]
Mulugruntz: Oh, you're right. Then this becomes out of scope. I'll resolve this. Maybe we can create a ticket for new-developer, WDYT?

myhoard/web_server.py
```diff
@@ -201,6 +239,8 @@ def _add_routes(self):
             [
                 web.get("/backup", self.backup_list),
                 web.post("/backup", self.backup_create),
+                web.post("/backup_cancel_preserve", self.backup_cancel_preserve),
+                web.post("/backup_preserve", self.backup_preserve),
```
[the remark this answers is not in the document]
kathia-barahona: Just following the same format as pghoard, rather have a common pattern when actually implementing the preservation.

Anyways, most of the endpoint definitions we have is messy and misleading. For example, restoring a backup is `/PUT status` and sending the stream_id on the body... when ideally it should be `/backup/{stream_id}/restore`. So, its a bit arbitrary to decide which is the best way to expose a functionality through the API , 
refactoring this is not my main focus here :woman_shrugging:  (Even `GET /backup`, since it is not a plural, so wouldn't expect a list of backups )

I can change the method to PUT  (considered it initially but I overthinking since the endpoint only creates a new file, so it can also be POST, but anyways). 

We can focus on refactoring myhoard in other issues.

myhoard/controller.py
```diff
@@ -1087,10 +1104,17 @@ def _get_upload_backup_site(self):
...
+        if not backup:
+            KeyError(f"Stream {stream_id} not found in backups")
+        return backup["site"]
+
+    def _get_backup_by_stream_id(self, stream_id: str):
         for backup in self.state["backups"]:
             if backup["stream_id"] == stream_id:
-                return backup["site"]
-        raise KeyError(f"Stream {stream_id} not found in backups")
+                return backup
```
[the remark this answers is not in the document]
kathia-barahona: We do filter backups by stream_id lots of times... but that sounds as a major change and multiple places on the code will be affected, not sure if it will be worth it at the end of the day.

myhoard/web_server.py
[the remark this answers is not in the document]
Mulugruntz: Wow you're right, it is indeed messy. I didn't dig too deep into the other ones. But true, it may be worth considering having a ticket to have a `/v2/...` (keep backward compatibility), start deprecating old endpoints, and some time in the future (major myhoard version), remove them.

Agreed, see what you can do with the new ones to make them "not too messy at a low cost" and we'll see about some major thing in the future for this.

myhoard/controller.py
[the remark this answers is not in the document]
Mulugruntz: Okay, out of scope then. 👍 
At the end of the review, I'll create a few tickets about all the points mentioned, to keep track of them and they can be prioritized by Amichai at a later date.

myhoard/web_server.py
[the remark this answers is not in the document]
rikonen: I think the API suggested by @Mulugruntz is better.

`/status` is different because that endpoint updates the global status of what the application is doing. You cannot be restoring multiple different backups at the same time and having an API that says you want to update status to restoring backup and give the backup as parameter is imo clearer than having the restoration as a property of a backup. The restoration may also automatically switch to restoring a different backup if the requested one was broken.

For backup preservation this is clearly a property of an individual backup and any number of the backups may be preserved.

myhoard/web_server.py
```diff
@@ -78,6 +80,42 @@ async def backup_list(self, _request):
...
+            stream_id = body.get("stream_id")
+            if stream_id is None:
+                raise BadRequest("`stream_id` must be set in request body.")
+
+            return self._handle_preservation_request(PreservationRequest(stream_id=stream_id, preserve_until=None))
+
+    def _handle_preservation_request(self, preservation_request: PreservationRequest):
+        try:
+            with self.controller.lock:
+                self.controller.mark_backup_preservation(preservation_request)
```
rikonen, line 114: This blocks the asyncio thread and makes the API non-functional. We should not be doing any blocking operations here. Also holding the controller lock throughout the entire operation is bad, the lock is intended to only be held while checking what current state is and updating the state, not while performing network IO. Presumably this should just update the controller state to indicate preservation request has been issued and then wait for the operation to completely like we do elsewhere.

rikonen (dismissed):

myhoard/web_server.py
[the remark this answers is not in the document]
rikonen: Or rather than having the separate `/preserve` endpoint I think we should just have `PUT /backup/<stream_id>` operation where the body can contain `{"preserve_until": "<timestamp>"}` value.

alexole (approved):

alexole (dismissed):
Clicked a wrong button

myhoard/web_server.py
kathia-barahona: :thinking: Though shouldn't the controller be locked in order to avoid backup deletion while updating the preservation file? Also, requesting preservation should refresh the backup list.

kathia-barahona: Tho I am noticing we are actually not consistent with this :/ Seems we acquire the controller lock a couple of times in the API (e.g `backup_list`)  and we are not acquiring the lock when calling `_refresh_backups_list`... it updates the controller state without acquiring the lock. Unless I am understanding wrong your comment :)

rikonen: For backup deletion `_purge_old_backups` should just check the pending preservation requests and not delete a backup that has been preserved.

Looks like `_refresh_backup_list()` is missing lock around the `self.state_manager.update_state(backups=backups, backups_fetched_at=time.time())` call but I don't think it needs anything beyond that.

myhoard/web_server.py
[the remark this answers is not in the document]
rikonen: Did you consider the `PUT /backup/<stream_id>` (or perhaps more appropriately `PATCH /backup/<stream_id>`)? I don't see any obvious downside with that and it would be much more restful approach.

myhoard/web_server.py
kathia-barahona: yes, though `_purge_old_backups` is removing the backup stream without acquiring the lock (it later acquires it for updating owned stream ids). So we might reach a race condition when trying to preserve it... stream is still valid but at the moment of preservation it suddenly got removed, which might lead to a false positive.

I'll open a separate PR for `_refresh_backup_list` lock acquisition, that should remove the lock acquisition in `GET /backup`

myhoard/controller.py
```diff
@@ -1408,11 +1434,12 @@ def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
             self.stats.gauge_float("myhoard.binlog.time_since_any_purged", current_time - last_purge)
             self.stats.gauge_float("myhoard.binlog.time_since_could_have_purged", current_time - last_could_have_purged)
 
-    def _refresh_backups_list(self):
+    def _refresh_backups_list(self, force_refresh: bool = False):
```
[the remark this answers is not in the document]
kathia-barahona: this requires a lot of changes since `Controller.get_backup_list` is not returning `List[Backup]`, is returning `List[Dict[.....]]`. I rather keep this PR small and not involve any refactoring work that is not required. I'll create an issue for this.

myhoard/backup_stream.py
```diff
@@ -437,6 +439,30 @@ def mark_as_closed(self) -> None:
         )
         self.wakeup_event.set()
 
+    def mark_preservation(self, preserve_until: Optional[datetime]) -> None:
+        self.log.info("Marking stream %s preservation.", self.stream_id)
```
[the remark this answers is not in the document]
kathia-barahona: yeeh, wanted to avoid an if/else just for a log message since it is optional. But will do

myhoard/web_server.py
rikonen: > though _purge_old_backups is removing the backup stream without acquiring the lock

We can make that update some local state to indicate that we're already in the process of removing a backup and if a preservation request is done for a backup we're already deleting, the request fails.

myhoard/web_server.py
```diff
@@ -78,6 +80,42 @@ async def backup_list(self, _request):
...
+                return json_response({"success": True})
+        except Exception as ex:
+            raise BadRequest(str(ex))
```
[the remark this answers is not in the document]
kathia-barahona: Did `PUT /backup/{stream_id}/preserve` instead. And now it returns 500 in case there was an internal error (something wrong when writing the file for example)

myhoard/backup_stream.py
```diff
@@ -437,6 +439,30 @@ def mark_as_closed(self) -> None:
...
+
+        preserved_info = {
+            "preserve_until": preserve_until.isoformat() if preserve_until else None,
```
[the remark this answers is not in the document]
kathia-barahona: It does now:
 https://github.com/Aiven-Open/myhoard/blob/74a876548651375ed414d4cb374255e234f49c61/myhoard/web_server.py#L82-L88

myhoard/controller.py
```diff
@@ -1304,8 +1334,15 @@ def _purge_old_backups(self):
...
-        self._build_backup_stream(backup).remove()
+        # lock the controller, this way other requests do not access backups till backup is purged
         with self.lock:
+            self._build_backup_stream(backup).remove()
+            current_backups = [
+                current_backup
+                for current_backup in self.state["backups"]
+                if current_backup["stream_id"] != backup["stream_id"]
+            ]
+            self.state_manager.update_state(backups=current_backups)
```
kathia-barahona, line 1345: Updated list of backups here, otherwise the purged backup might still be listed (_refresh_backup_list might skip refreshing in the next iteration)

myhoard/controller.py
```diff
@@ -265,6 +266,17 @@ def mark_backup_requested(
...
+    def mark_backup_preservation(self, stream_id: str, preserve_until: Optional[datetime.datetime]) -> None:
+        backup_to_preserve = self._get_backup_by_stream_id(stream_id)
+        if not backup_to_preserve:
+            raise Exception(f"Stream {stream_id} was not found in completed backups.")
+
+        # should acquire lock in case the controller tries to purge the backup
+        with self.lock:
+            self._build_backup_stream(backup_to_preserve).mark_preservation(preserve_until=preserve_until)
+            # force refresh, this way we guarantee preserve_until is updated and backup does not get deleted
+            self._refresh_backups_list(force_refresh=True)
```
rikonen, line 278: No network IO must be done while holding the lock and this is also blocking network call from asyncio context which blocks the API.

kathia-barahona: aff I forgot to remove that after previous clarification. Thanks!

myhoard/controller.py
```diff
@@ -1087,10 +1104,17 @@ def _get_upload_backup_site(self):
         return site_id, non_recovery_sites[site_id]
 
     def _get_site_for_stream_id(self, stream_id):
```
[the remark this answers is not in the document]
kathia-barahona: can add the `stream_id: str` but not the return type :disappointed:  List of backups in the state is not typed correctly, so lots of changes

myhoard/web_server.py
```diff
@@ -78,6 +79,24 @@ async def backup_list(self, _request):
...
+            if preserve_until is not None:
+                try:
+                    preserve_until = datetime.fromisoformat(preserve_until).astimezone(timezone.utc)
+                    now = datetime.now(timezone.utc)
+                    if preserve_until < now:
+                        raise BadRequest("`preserve_until` must be a date in the future.")
+                except ValueError:
+                    raise BadRequest("`preserve_until` must be a valid isoformat datetime string.")
+
+            self.controller.mark_backup_preservation(stream_id=stream_id, preserve_until=preserve_until)
```
rikonen, line 97: Maybe some misunderstanding here but this is still a blocking call that makes blocking network IO and we must not do that from async context as the API will be blocked. We should instead update local state to indicate that backup is expected to be preserved and then use async wait here to wait for the request to actually go through.

kathia-barahona: hah that's true... maybe having a queue for backup preservation requests and later make the controller handle it?  Because the request might take a bit of time (depending how long it takes refreshing backups), so maybe handling the batch of requests and later refreshing might be ideal. I just find a bit misleading returning "success"

rikonen: Just having some `{"pending_preservation_requests": {"<stream_id_1>": "<timestamp_1>"|null, "<stream_id_2>": "<timestamp_2>"|null}}` would presumably be quite sufficient way of storing this information in controller state. It could then drop those entries from `pending_preservation_requests` once the data has actually been updated.

I think the `while True: await asyncio.sleep(1); check_if_operation_is_done_yet` pattern used in some other handler should be OK for this. If the operation takes a very long time, like over 60 seconds, this could raise timeout error to avoid returning misleading response.

kathia-barahona: well... maybe not even a queue, just updating the state is enough

kathia-barahona: :+1:  alright, will add `wait_for_applied_preservation` as optional in the request's body and wait only in this case (we do something similar for backup_create)

myhoard/web_server.py
```diff
@@ -78,6 +79,59 @@ async def backup_list(self, _request):
...
+                        break
+
+                    elapsed = time.monotonic() - start
+                    if elapsed > wait_for_applied_preservation:
+                        self.log.info(
+                            "Preservation for backup %s was not applied up in %.1f seconds",
+                            stream_id,
+                            elapsed,
+                        )
+                        break
```
rikonen, line 128: I think this case should rather raise an error than return success, otherwise the timeout is rather meaningless as caller cannot anyway know what happened.

myhoard/web_server.py
```diff
@@ -78,6 +79,59 @@ async def backup_list(self, _request):
...
+            wait_for_applied_preservation = body.get("wait_for_applied_preservation")
+            if wait_for_applied_preservation:
+                self.log.info(
+                    "Waiting up to %.1f seconds for preservation of backup %s to be applied.",
+                    wait_for_applied_preservation,
+                    stream_id,
+                )
+                start = time.monotonic()
+                while True:
+                    backup = self.controller.get_backup_by_stream_id(stream_id)
```
rikonen, line 107: Doesn't look like a public method named this would exist?

myhoard/controller.py
```diff
@@ -265,6 +268,18 @@ def mark_backup_requested(
...
             self.state_manager.update_state(backup_request=new_request)
 
+    def mark_backup_preservation(self, stream_id: str, preserve_until: Optional[datetime.datetime]) -> None:
+        backup_to_preserve = self._get_backup_by_stream_id(stream_id)
+        if not backup_to_preserve:
+            raise Exception(f"Stream {stream_id} was not found in completed backups.")
+
+        with self.lock:
+            current_requests = self.state["pending_preservation_requests"]
+            current_requests[stream_id] = preserve_until.isoformat() if preserve_until else None
```
rikonen, line 278: Afair there's a gotcha with how the state update works as it compares current state against previously persisted state and if you modify the state in-place it doesn't see the change and won't persist it on disk. In-place upgrades also have more potential for introducing concurrency issues. This needs to instead be written as
```
new_requests = dict(self.state["pending_preservation_requests"])
new_requests[stream_id] = preserve_until.isoformat() if preserve_until else None
```
so that a copy of the streams is updated instead of the original one.

rikonen (dismissed):
This has quite a lot of commits that look like internal development history, didn't check if there are some that make sense as individual ones but probably they should mostly be squashed.

myhoard/controller.py
```diff
@@ -1264,7 +1290,10 @@ def _process_removed_binlogs(self, binlogs):
                 stream.remove_binlogs(binlogs)
 
     def _purge_old_backups(self):
-        purgeable = [backup for backup in self.state["backups"] if backup["completed_at"]]
+        # wait for acquiring lock in case the controller got a request for refreshing backups.
+        # this way preserve_until and broken_at attributes are up to date before purging
+        with self.lock:
```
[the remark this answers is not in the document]
kathia-barahona: ah true, this lock doesn't make sense anymore after I changed the way of marking preservation. Was more intended in case there was a race condition... but no concern anymore regards that.

myhoard/web_server.py
kathia-barahona: should we raise the error also for backup_create? https://github.com/Aiven-Open/myhoard/blob/965421770f091ff1d3dc05ba4ddf3486a6fed0b9/myhoard/web_server.py#L55-L67  I was just following the current pattern

myhoard/controller.py
```diff
@@ -1431,6 +1473,22 @@ def _stream_for_backup_initiated_by_old_master(self, stream: BackupStream) -> bo
...
+            return None
+
+        for stream_id, preserve_until in self.state["pending_preservation_requests"].items():
+            backup_to_preserve = self._get_backup_by_stream_id(stream_id)
+            if backup_to_preserve:
+                self._build_backup_stream(backup_to_preserve).mark_preservation(
+                    preserve_until=datetime.datetime.fromisoformat(preserve_until) if preserve_until else None
+                )
+
+        with self.lock:
```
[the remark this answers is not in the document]
kathia-barahona: Reasonable. Thanks a lot for the comments!! Must admit I didn't paid much attention to the state files and network operations before this task, will keep in mind  :smile:

myhoard/web_server.py
kathia-barahona: apparently Pycharm's refactor played against.

myhoard/web_server.py
rikonen: Ah, I didn't realize it worked like this. I don't remember why it was done like that, seems weird looking at it now. I guess consistency is good but maybe then at least add some flag in the response to indicate the operation is still ongoing.

myhoard/controller.py
kathia-barahona: Updated: https://github.com/Aiven-Open/myhoard/blob/02605fbb08f5d6cfa3ff34aa34e852d81b4d2c95/myhoard/controller.py#L278-L281

myhoard/controller.py
```diff
@@ -1304,8 +1338,15 @@ def _purge_old_backups(self):
             self.log.warning("Backup %r to drop is one of active streams, not dropping", backup["stream_id"])
             return
 
-        self._build_backup_stream(backup).remove()
```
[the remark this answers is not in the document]
kathia-barahona: Changed, added `stream_to_be_purged` in the controller's state instead.

myhoard/web_server.py
kathia-barahona: should be public now

myhoard/controller.py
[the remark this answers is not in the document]
kathia-barahona: Opened https://github.com/Aiven-Open/myhoard/pull/175 for this

myhoard/controller.py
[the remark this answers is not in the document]
kathia-barahona: https://github.com/Aiven-Open/myhoard/pull/175 handles this

rikonen (approved):
Leaving still open for the team working on this component nowadays to review but as far as I'm concerned looks good now.

One gotcha this will introduce is that if a backup is preserved for so long that we end up purging a backup that was newer than the one we preserved, we end up with a hole in the stream of binlogs and the mechanism that automatically restores an older backup plus binlogs of all newer backups if we encounter a broken backup will not work. I suspect that won't be an issue in practice as this is presumably just a temporary measure we can use to keep backups for some extra minutes or hours.

alexole (approved):

PR #172, opened 2023-09-21
acquire lock when updating list of backups

kathia-barahona:
acquire lock when updating list of backups

Controller should acquire lock every time the status is being updated.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1422,7 +1422,9 @@ class Controller(threading.Thread):
         for backup in self.state["backups"]:
             if backup["stream_id"] not in new_backups_ids:
                 self._delete_backup_stream_state(backup["stream_id"])
-        self.state_manager.update_state(backups=backups, backups_fetched_at=time.time())
+
+        with self.lock:
+            self.state_manager.update_state(backups=backups, backups_fetched_at=time.time())
         return backups
 
     def _refresh_backups_list_and_streams(self):


myhoard/web_server.py
```diff
@@ -73,10 +73,9 @@ async def backup_list(self, _request):
             response = {
                 "backups": None,
             }
-            with self.controller.lock:
```
rikonen, line 76: The change for `controller` is valid but why is the lock removed here? We need both for the access to be properly protected.

kathia-barahona: mmm I think I am a bit confused since we do not apply acquire the lock when accessing to backups, so not sure why we should do here. Also, maybe we should avoid accessing to the controller's state from other places?  Is it better to have `Controller.get_backups()` and lock it from there?

rikonen: So that both the `backups_fetched_at` and `backups` are accessed in a consistent state. Otherwise a statement that updates these values could've already set `backups_fetched_at` to some value and then we assume here backups are available but then that value hasn't yet been set so we could end up returning empty backup list instead of `null`.

Controller itself doesn't need the lock because it is the only one that is making updates to the backups and the main logic there runs on the same thread so there cannot be concurrent update.

kathia-barahona: ah fair enough, thanks for the explanation :+1:  will bring it back

rikonen (approved):

PR #173, opened 2023-09-21
Mark streams opened for old master's backups as closed [BF-2225]

ettanany:
### About this change: What it does, why it matters

When the role of the current node is master, then we mark any observe streams as closed, as those are assumed to be opened for backups initiated by the old master after the promotion process of the current node was triggered. This is necessary so that new backups can be requested, as we do not allow creating new streams if 2 or more are already opened.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -1425,6 +1425,11 @@ class Controller(threading.Thread):
         self.state_manager.update_state(backups=backups, backups_fetched_at=time.time())
         return backups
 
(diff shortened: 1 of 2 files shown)


rikonen (dismissed):
Seems reasonable. The promotion logic should refresh the list of backup streams and promote any streams in observe mode into active mode and we should not end up in a situation where mode is active yet streams are in observe mode except due to some race condition.

alexole (approved):

PR #174, opened 2023-09-22
track block mismatch errors when taking a basebackup

kathia-barahona:

Sometimes xtrabackup might fail when keeping up with redo logs, this will end up with an error due to block mismatches. Will be nice to track such errors, this way maybe we know when to set `redo_log_archiving=true` and retry taking the backup again... as just restarting myhoard is not the proper solution.

Seems there is a misconception that xtrabackup is not releasing correctly the backup lock, when the case is that the BackupStream keeps retrying on taking a basebackup, xtrabackup takes the lock and eventually fails due to block mismatch. This might take a long time, so myhoard will keep complaining when purging binlogs.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -2,7 +2,7 @@
 from .append_only_state_manager import AppendOnlyStateManager
 from .basebackup_operation import BasebackupOperation
 from .binlog_scanner import BinlogInfo
-from .errors import XtraBackupError
+from .errors import BlockMismatchError, XtraBackupError
 from .state_manager import StateManager
 from .util import (
     add_gtid_ranges_to_executed_set,
@@ -67,6 +67,12 @@ class PromotionInfo(TypedDict):
     promoted_at: float
     server_id: int
     start_index: int
+
+
+class BaseBackupFailureReason(str, enum.Enum):
+    network_error = "network_error"
+    block_mismatch_error = "block_mismatch_error"
+    xtrabackup_error = "xtrabackup_error"
 
 
 class BackupStream(threading.Thread):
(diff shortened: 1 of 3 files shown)


myhoard/backup_stream.py
```diff
@@ -69,6 +69,17 @@ class PromotionInfo(TypedDict):
...
 
+class BaseBackupFailureReason(str, enum.Enum):
+    network_error = "network_error"
+    block_mismatch_error = "block_mismatch_error"
+    xtrabackup_error = "xtrabackup_error"
+
+
+class BaseBackupAttempt(TypedDict):
+    failed_at: float
+    reason: BaseBackupFailureReason
```
kathia-barahona, line 80: **comment** I might remove this, maybe keeping the reason of the failures might help the BackupStream to know what to do on each particular case, but still not sure about it

myhoard/basebackup_operation.py
```diff
@@ -288,6 +289,8 @@ def _process_input_output(self):
 
         if exit_code != 0:
             self.log.error("xtrabackup exited with non-zero exit code %s: %r", exit_code, pending_output)
+            if self.has_block_mismatch:
+                raise BlockMismatchError(f"xtrabackup failed with code {exit_code} due log block mismatch.")
```
kathia-barahona, line 293: **comment** Had to track separately the block mismatch error from other output line, xtrabackup will finish with a different output that those not indicates this as root cause.

myhoard/basebackup_operation.py
```diff
@@ -310,6 +313,9 @@ def _process_output_line(self, line):
             and not self._process_output_line_file_finished(line)
             and not self._process_output_line_lsn_info(line)
         ):
+            if "log block numbers mismatch" in line:
+                self.has_block_mismatch = True
```
alexole, line 317: nit: probably does not make much sense to do it here, we could save some CPU pennies if we move it to the next block where we know it's an error.

myhoard/backup_stream.py
```diff
@@ -1021,13 +1027,24 @@ def _take_basebackup(self) -> None:
         ) as ex:
             self.log.exception("Network error while taking basebackup")
             self.state_manager.increment_counter(name="basebackup_errors")
-            self.stats.increase("myhoard.basebackup.errors", tags={"ex": ex.__class__.__name__, "reason": "network_error"})
+            self.stats.increase(
+                "myhoard.basebackup.errors",
+                tags={"ex": ex.__class__.__name__, "reason": BaseBackupFailureReason.network_error},
+            )
             self.last_basebackup_attempt = time.monotonic()
-        except Exception as ex:  # pylint: disable=broad-except
```
alexole, line 1026: Are we sure no other exception can be thrown here except for XtraBackupError?

kathia-barahona: hmmm maybe :thinking: but then we are not tracking `myhoard.basebackup.errors` correctly... or at least it is a bit misleading. But let me test, I think we raise `AbortRequested` exceptions

myhoard/backup_stream.py
```diff
@@ -1021,13 +1027,24 @@ def _take_basebackup(self) -> None:
...
+            )
             self.last_basebackup_attempt = time.monotonic()
         except Exception as ex:  # pylint: disable=broad-except
             self.log.exception("Failed to take basebackup")
-            self.stats.increase("myhoard.basebackup.errors", tags={"ex": ex.__class__.__name__})
+
+            # explicitly tag block mismatch errors, this way we can relate high redo log activity to xtrabackup errors
+            if isinstance(ex, BlockMismatchError):
+                failure_reason = BaseBackupFailureReason.block_mismatch_error
+            else:
```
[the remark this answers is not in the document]
alexole: And if it's not it should be some other default reason or no reason.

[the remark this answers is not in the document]
kathia-barahona: true

alexole (approved):

PR #175, opened 2023-09-28
use dictionary instead of list for Controller.state['backups']

kathia-barahona:

Changed the way we store backups in the controller's state, main reason is because there a lot of scenarios where we need to fetch a backup based on its stream id.  Also added more typing when backups are refreshed.

Not super urgent, but will be nice to have.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -75,6 +75,26 @@ class BaseBackupFailureReason(str, enum.Enum):
     xtrabackup_error = "xtrabackup_error"
 
 
+class BaseBackup(TypedDict, total=False):
+    binlog_index: Optional[int]
+    binlog_name: Optional[str]
+    binlog_position: Optional[int]
+    backup_reason: Optional["BackupStream.BackupReason"]
+    compressed_size: float
+    encryption_key: bool
+    end_size: Optional[int]
+    end_ts: float
+    gtid: str
+    gtid_executed: Dict[str, List[List[int]]]
+    initiated_at: float
+    lsn_info: Optional[Dict[str, float]]
+    normalized_backup_time: Optional[str]
+    number_of_files: int
+    start_size: Optional[int]
+    start_ts: float
+    uploaded_from: int
+
+
 class BackupStream(threading.Thread):
     """Handles creating a single consistent backup stream. 'stream' here refers to uninterrupted sequence
     of backup data that can be used to restore the system to a consistent state. It includes the basebackup,
@@ -121,7 +141,7 @@ class BackupStream(threading.Thread):
         backup_errors: int
         basebackup_errors: int
         basebackup_file_metadata: Optional[Dict]
-        basebackup_info: Dict
+        basebackup_info: BaseBackup
         broken_info: Dict
         closed_info: Dict
         completed_info: Dict
@@ -1007,7 +1027,7 @@ class BackupStream(threading.Thread):
                 self.log.info("Last basebackup GTID %r, truncating GTID executed %r accordingly", last_gtid, gtid_executed)
                 truncate_gtid_executed(gtid_executed, last_gtid)
 
-            info = {
+            info: BaseBackup = {
                 "binlog_index": int(binlog_info["file_name"].split(".")[-1]) if binlog_info else None,
                 "binlog_name": binlog_info["file_name"] if binlog_info else None,
                 "binlog_position": binlog_info["file_position"] if binlog_info else None,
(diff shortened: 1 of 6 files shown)


myhoard/backup_stream.py
```diff
@@ -75,6 +75,26 @@ class BaseBackupFailureReason(str, enum.Enum):
     xtrabackup_error = "xtrabackup_error"
 
 
+class BaseBackup(TypedDict, total=False):
```
kathia-barahona, line 78: Just moved this from myhoard.Controller and added rest of expected fields.  The `total=False` is mainly in case a `basebackup.json` file is not containing all keys (ideally shouldn't happen but just in case)

myhoard/controller.py
```diff
@@ -548,60 +542,54 @@ def collect_binlogs_to_purge(
...
-    @staticmethod
-    def get_backup_list(backup_sites: Dict[str, BackupSiteInfo], *, seen_basebackup_infos=None, site_transfers=None):
-        if seen_basebackup_infos is None:
-            seen_basebackup_infos = {}
-        if site_transfers is None:
-            site_transfers = {}
-        backups = []
-        for site_name, site_config in backup_sites.items():
-            file_storage = site_transfers.get(site_name)
+    def get_backups(self) -> Dict[str, Backup]:
```
kathia-barahona, line 545: Removed previous arguments... :woman_shrugging: we were just passing class properties that can be accessed here and it was quite odd.

myhoard/controller.py
```diff
@@ -548,60 +542,54 @@ def collect_binlogs_to_purge(
...
                         broken_info = parse_fs_metadata(info["metadata"])
                     elif file_name == "closed.json":
                         closed_info = parse_fs_metadata(info["metadata"])
                     elif file_name == "completed.json":
                         completed_info = parse_fs_metadata(info["metadata"])
 
-                if basebackup_info and basebackup_compressed_size:
-                    basebackup_info = dict(basebackup_info, compressed_size=basebackup_compressed_size)
+                if basebackup_info and basebackup_compressed_size is not None:
+                    basebackup_info["compressed_size"] = basebackup_compressed_size
```
kathia-barahona, line 577: **question** Shouldn't `basebackup.json` already contain the compressed size? This looked quite weird tbh, left it as it is since I am just focusing on typing rather than behavior

myhoard/web_server.py
```diff
@@ -75,7 +75,7 @@ async def backup_list(self, _request):
             }
             with self.controller.lock:
                 if self.controller.state["backups_fetched_at"]:
-                    response["backups"] = self.controller.state["backups"]
+                    response["backups"] = list(self.controller.state["backups"].values())
```
kathia-barahona, line 78: **comment** left as a list here, I don't see a reason for changing the response type

myhoard/controller.py
rikonen: Looks like it does. I checked the private MyHoard repo with development history and it was not there at one point during development, this is probably a leftover from those times that is not needed because in any officially released version the field is always present.

myhoard/controller.py
```diff
@@ -1086,10 +1074,9 @@ def _get_upload_backup_site(self):
             raise Exception("Defined upload site not present in list of non-recovery backup sites")
         return site_id, non_recovery_sites[site_id]
 
-    def _get_site_for_stream_id(self, stream_id):
-        for backup in self.state["backups"]:
-            if backup["stream_id"] == stream_id:
-                return backup["site"]
+    def _get_site_for_stream_id(self, stream_id: str) -> str:
+        if stream_id in self.state["backups"]:
+            return self.state["backups"][stream_id]["site"]
```
rikonen, line 1079: I didn't check what threads this is accessed from but probably to avoid any future problems it would anyway be better to put this if within a `with self.lock:` block because access of `"backups"` is no longer atomic with this change.

rikonen:
I don't know if we have any policy related to backward compatibility and probably earlier changes have been incompatible too (perhaps to lesser extent) but if anyone is doing in-place upgrades for MyHoard then it will break when upgrading from a version prior to this one.

kathia-barahona:
> I don't know if we have any policy related to backward compatibility and probably earlier changes have been incompatible too (perhaps to lesser extent) but if anyone is doing in-place upgrades for MyHoard then it will break when upgrading from a version prior to this one.

Oh yes, this will definitely break for in-place upgrades :grimacing: tho not sure if working on backward compatibility is worth it, old state file should be just removed

PR #176, opened 2023-10-06
remove backup preservation after restoration

kathia-barahona:
Remove preservation after the backup is restored. This will help avoiding preserving a backup for too long when it is just locked for restoring

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,6 +552,7 @@ Create a request for updating the preservation status of a backup. Request body 
 ```
 {
   "preserve_until": "2023-09-01T00:00:0000",
+  "remove_after_restore": true,
   "wait_for_applied_preservation": 3.0,
 }
 ```
@@ -565,6 +566,11 @@ Identifier of this backup.
 Optional datetime value in ISO format for keeping the backup from being deleted.
 If a valid value is provided, the backup will be preserved until the specified datetime.
 If not provided or has a null value, the backup can be deleted due to old age.
+
+
+**remove_after_restore**
+
+If true, preservation will be removed after the backup is restored.
 
 **wait_for_applied_preservation**
 
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -445,7 +445,7 @@ class BackupStream(threading.Thread):
         )
         self.wakeup_event.set()
 
-    def mark_preservation(self, preserve_until: Optional[datetime]) -> None:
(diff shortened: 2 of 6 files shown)


rikonen:
I'm not sure how we're planning to use this functionality exactly but you could fork N new services from the same old backup and the backup should be preserved until all forks are up, which sounds like it might go against the assumption here that backup is safe to remove once it has been restored.

kathia-barahona:
> I'm not sure how we're planning to use this functionality exactly but you could fork N new services from the same old backup and the backup should be preserved until all forks are up, which sounds like it might go against the assumption here that backup is safe to remove once it has been restored.

Yes, thought about that. More of an experiment here tbh, @rikonen do you think we should keep track of a list service ids that have an ongoing restoration of the backup in the object storage? I considered that, but feels like I am getting into a rabbit hole... 
The main idea is to avoid preserving backups for too long, since currently the master has no way to know when the forks have finished restoring (well, actually yes if we check the standby/fork nodes statuses, but how to know if the backup was meant to be preserved due to restoration?)

rikonen:
I don't think this should be the responsibility of MyHoard at all. In general having new services that start by reusing backups of an existing service modify those backups seems bad, the backups should be considered read-only and not touched. The general management logic should determine when backups need to be retained and when removing those is safe. E.g. the configuration we send to the source service could have a list of new services being forked from that service and the backups used for doing the forking, which the control logic of the source service can then make appropriate request to MyHoard to start preserving a backup when there's an ongoing restoration for it and stop preserving it when last of those goes away.

kathia-barahona:
> I don't think this should be the responsibility of MyHoard at all. In general having new services that start by reusing backups of an existing service modify those backups seems bad, the backups should be considered read-only and not touched. The general management logic should determine when backups need to be retained and when removing those is safe. E.g. the configuration we send to the source service could have a list of new services being forked from that service and the backups used for doing the forking, which the control logic of the source service can then make appropriate request to MyHoard to start preserving a backup when there's an ongoing restoration for it and stop preserving it when last of those goes away.

I agree, then this means we should remove such mechanism from pghoard as well... It has a flag for removing preservation after restoration, tho the fork is the one actually requesting the preservation (master is not aware) before restore. Still dangerous IMHO since the backup can still be removed before the node reaches a setup stage.

Will close this PR and create an issue for changing behavior in pghoard as well.

PR #180, opened 2024-01-16
Make restore workers pool configurable

kmichel-aiven:

Starting as many Python interpreters as there are CPUs can be problematic, particularly when running tests using myhoard and when these tests are also parallelized.


alexole (approved):
Thanks for the contribution 🎉

PR #182, opened 2024-03-04
Use rohmu transfer pool to fix io error on closed fds on restore

giacomo-alzetta-aiven:

Currently myhoard can fail with: `ValueError: I/O operation on closed file` during restore (. The problem is likely due to non thread-safety of the rohmu transfers.

Using rohmu's transfer pool we should eliminate any possible case in which multiple threads operated on the same transfer object, or cases in which a "stale" transfer object was used by a single thread.


giacomo-alzetta-aiven:
Seem like CI is broken? It failed to setup the linting...

alexole (approved):

PR #184, opened 2024-03-04
Restrict env update command to specific variable and values

0xlianhu:
Restrict/simplify`myhoard_mysql_env_update` script can only update the `MYSQLD_OPTS` variable with specific values.

### DIFF
--- a/myhoard/myhoard.py
+++ b/myhoard/myhoard.py
@@ -113,19 +113,18 @@ class MyHoard:
         self.systemd_notified = True
 
     def _restart_mysqld(self, *, with_binlog, with_gtids):
-        mysqld_options = []
-        if not with_binlog:
-            mysqld_options.append("--disable-log-bin")
-            # If config says slave-preserve-commit-order=ON MySQL would refuse to start if binlog is
-            # disabled. To prevent that from happening ensure preserve commit order is disabled
-            mysqld_options.append("--skip-slave-preserve-commit-order")
-        if not with_gtids:
-            mysqld_options.append("--gtid-mode=OFF")
-
         systemd_service = self.config.get("systemd_service")
         if systemd_service:
-            self._restart_systemd(mysqld_options=" ".join(mysqld_options), service=systemd_service)
+            self._restart_systemd(with_binlog=with_binlog, with_gtids=with_gtids, service=systemd_service)
         else:
+            mysqld_options = []
+            if not with_binlog:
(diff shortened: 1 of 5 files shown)


myhoard/myhoard.py
```diff
@@ -182,6 +172,21 @@ def _restart_systemd(self, *, mysqld_options, service):
...
+        try:
+            with open(env_file_path, "r") as f:
+                contents = [
+                    line.rstrip("\n") for line in f.readlines() if line.strip() and not line.startswith(key)
+                ]
+        except FileNotFoundError:
+            contents = []
+        if value is not None:
+            contents.append(f"{key}={value}")
+        with atomic_create_file(env_file_path, binary=True, inherit_owner_from_parent=True) as f:
```
paulwouters, line 187: Creating a file from scratch would break selinux policies on the file. You either need to call restorecon on it, or better, you only modify the existing file.

0xlianhu: The `atomic_create_file` function has a bad name. The file should already exist.

myhoard/update_mysql_environment.py
```diff
@@ -13,26 +15,34 @@ def __init__(self, args):
...
+        # make sure we only update the parameter we need to update i.e., MYSQLD_OPTS
+        key = "MYSQLD_OPTS"  # we only update this environment variable
+        value = ""
+        if self.args.with_bin_log != "true":
+            value += "--disable-log-bin "
+            # If config says slave-preserve-commit-order=ON MySQL would refuse to start if binlog is
+            # disabled. To prevent that from happening ensure preserve commit order is disabled
+            value += "--skip-slave-preserve-commit-order "
+        if self.args.gtid_mode != "true":
+            value += "--gtid-mode=OFF"
```
rikonen, line 27: Creating a list of values and producing a string with `" ".join(values)` would be cleaner and avoid the extra whitespace that may sometimes appear in the output.

rikonen (commented):
Looks good to me except the value could in some cases contain extra trailing whitespace. I'd expect that to just get ignored but maybe it would be better to just construct the value the same way we used to as suggested in the other comment.

myhoard/update_mysql_environment.py
0xlianhu: Thanks, fixed.

rikonen (approved):

PR #190, opened 2024-10-29
Ensure MySQL is only restarted once before starting streaming changes

packi:
In some rare cases MySQL takes a long time to start accepting queries.
When that happens, the server may take several minutes to do some
initial processing, after which it starts responding normally again and
subsequent restarts are fast. Restarting it before it has finished the
processing causes the operation to never complete so avoid doing that
so that the system does not go to never ending restart loop.

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -1434,16 +1434,17 @@ class RestoreCoordinator(threading.Thread):
             return found, current_index
 
     def _ensure_mysql_server_is_started(self, *, with_binlog: bool, with_gtids: bool) -> None:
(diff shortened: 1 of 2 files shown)


rdunklau:
The restart case for systemd looks fine, as the commands  are checked, but I'm worried about the `_restart_process` case: the command output is not checked for, so a failure to start here will be silenced. I wouldn't matter with the previous code, as we would restart it anyway on the next try.

rdunklau:
After discussing with @packi the use without systemd is supposed to be for testing purposes only.

rdunklau (approved):

PR #193, opened 2024-11-14
Split basebackup files if configured

packi:

The configuration for this is on the storage site as the maximum
object-size varies between GCP/AWS and Azure.

[BF-2928]

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -3,6 +3,7 @@ from rohmu.errors import FileNotFoundFromStorageError
 
 from .append_only_state_manager import AppendOnlyStateManager
 from .basebackup_operation import BasebackupOperation
+from .binary_io_slice import BinaryIOSlice
 from .binlog_scanner import BinlogInfo
 from .errors import BlockMismatchError, XtraBackupError
 from .state_manager import StateManager
@@ -12,6 +13,7 @@ from .util import (
     DEFAULT_MYSQL_TIMEOUT,
     DEFAULT_XTRABACKUP_SETTINGS,
     ERR_TIMEOUT,
+    file_name_for_basebackup_split,
     first_contains_gtids_not_in_second,
     GtidExecuted,
     make_fs_metadata,
@@ -36,7 +38,6 @@ from socket import gaierror
 from socks import GeneralProxyError, ProxyConnectionError
 from ssl import SSLEOFError
 from typing import Any, Callable, cast, Dict, Iterable, Iterator, List, Optional, Set, Tuple, TYPE_CHECKING, TypedDict
-
 import contextlib
 import enum
 import json
@@ -151,6 +152,8 @@ class BackupStream(threading.Thread):
         stream_id: str
         updated_at: float
         valid_local_binlog_found: bool
+        number_of_splits: Optional[int]
+        split_size: Optional[int]
 
     def __init__(
         self,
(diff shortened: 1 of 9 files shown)


myhoard/binary_io_slice.py
```diff
@@ -0,0 +1,89 @@
...
+from typing import BinaryIO, Type, Iterable, Iterator
+
+
+class BinaryIOSlice(BinaryIO):
+    def __init__(self, max_file_size: int, stream: BinaryIO):
+        super().__init__()
+        self._size_remaining = max_file_size
+        self.stream = stream
+
+    def read(self, __n: int = ...) -> bytes:
```
jfg956, line 12: Performance Nerd Comment...

In below, I would avoif the `if` by introducing a new variable.  This new variable would be `m = min(__n, self._size_remaining)`.  I would then have a single call to `read`: `result = self.stream.read(m)`.

Some implementation of `min` avoids the `if`, so this might save some CPU cycles.  And IMHO this is cleaner code.

myhoard/binary_io_slice.py
```diff
@@ -0,0 +1,91 @@
...
+        to_read = min(__n, self._size_remaining)
```
alexole, line 17: I am wondering if it does what we want to in case if `__n = -1`, the `min` will return `-1` in this case, but we still want to read up to `self._size_remaining` ?

packi: I can just add add a check for >= 0 to solve that? Is -1 even supported by the IO classes?

packi: Seems to be supported, I'll probably just raise a `ValueError` in that case:
```
$ echo "asdf" > foo.txt
$ python
Python [...]
>>> with open("foo.txt", "r") as f:
...    print(f.read(-1))
... 
asdf

>>>
```

alexole: It should mean read until EOF if I am not mistaken.

packi: Rohmu will not use that but I've changed it to this, seems to be what a normal `read` does:
```
        if __n < 0:
            to_read = self._size_remaining
        else:
            to_read = min(__n, self._size_remaining)
```

myhoard/backup_stream.py
```diff
@@ -675,10 +683,45 @@ def upload_progress(bytes_sent):
                 stats=self.stats,
             )
 
-        file_storage.store_file_object(
-            self._build_full_name("basebackup.xbstream"), stream, metadata=metadata, upload_progress_fn=upload_progress
-        )
+        split_nr = 0
+        uploaded_size = 0
+        basebackup_file_path = self._build_full_name("basebackup.xbstream")
+        while (split_nr == 0) or uploaded_size == self.split_size:
```
alexole, line 689: If the last piece is exactly `self.split_size`, would it not cause any issues? It looks like it will still make an iteration over empty stream.

packi: We'll just upload an empty split-file which is fine. We could delete it but not sure we would want to add that complexity.
We'll also exit cleanly in case we're not splitting and the basebackup would be 0 bytes (for whatever reason that could happen), because of the `break` in the end. If that wouldn't be there we would have an issue.

alexole: Just wanted to make sure this case is covered and tested.

alexole (approved):

PR #197, opened 2025-01-16
Not marking backups as broken on version upgrade

egor-voynov-aiven:
If a backup fails to restore, but the Percona PXB version used to create the backup differs from the current version, we do not mark the backup as "broken". The issue may be related to Percona PXB version compatibility.

### DIFF
--- a/myhoard/basebackup_restore_operation.py
+++ b/myhoard/basebackup_restore_operation.py
@@ -1,9 +1,9 @@
 # Copyright (c) 2019 Aiven, Helsinki, Finland. https://aiven.io/
 from .errors import DiskFullError
-from .util import get_xtrabackup_version
+from .util import get_xtrabackup_version, parse_version, parse_xtrabackup_info
 from contextlib import suppress
 from rohmu.util import increase_pipe_capacity, set_stream_nonblocking
-from typing import Final, Optional
+from typing import Final, Optional, Tuple
 
 import base64
 import fnmatch
@@ -56,6 +56,7 @@ class BasebackupRestoreOperation:
         self.stream_handler = stream_handler
         self.temp_dir = None
         self.temp_dir_base = temp_dir
+        self.backup_xtrabackup_info = None
 
     def restore_backup(self):
         if os.path.exists(self.mysql_data_directory):
@@ -101,6 +102,16 @@ class BasebackupRestoreOperation:
                         self._process_xbstream_input_output()
 
                 self.data_directory_size_start = self._get_directory_size(self.temp_dir)
+                xtrabackup_info_path = os.path.join(self.temp_dir, "xtrabackup_info")
+                if os.path.exists(xtrabackup_info_path):
+                    with open(xtrabackup_info_path) as fh:
+                        xtrabackup_info_text = fh.read()
+                    self.backup_xtrabackup_info = parse_xtrabackup_info(xtrabackup_info_text)
+                    self.log.info(
+                        "Backup info. Tool version: %s, Server version: %s",
+                        self.backup_xtrabackup_info.get("tool_version"),
+                        self.backup_xtrabackup_info.get("server_version"),
+                    )
 
                 # TODO: Get some execution time numbers with non-trivial data sets for --prepare
                 # and --move-back commands and add progress monitoring if necessary (and feasible)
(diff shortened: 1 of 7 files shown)


myhoard/basebackup_restore_operation.py
```diff
@@ -56,11 +56,15 @@ def __init__(
         self.stream_handler = stream_handler
         self.temp_dir = None
         self.temp_dir_base = temp_dir
+        self.backup_xtrabackup_info = None
+        self.xtrabackup_version = None
```
ettanany, line 60: I think this field is not necessary. The reason I questioned its usage is that it wasn't clear to me in the beginning that it means the installed version of XtraBackup tool. But it looks like the only place where it's used is in `xtrabackup_version_match` function/property, where it's read from `state`, but do we even need to store it in state? We can always get it from the system by calling `get_xtrabackup_version` anywhere in the code (especially that it's only used once). In this case we can call it inside `xtrabackup_version_match`, which would significantly simplify the whole thing.

`xtrabackup_version_match` would end up looking something like below:

```diff
     @property
-    def xtrabackup_version_match(self) -> bool | None:
-        if self.state["xtrabackup_version"] is None or self.state["backup_xtrabackup_version"] is None:
-            return None
-        return self.state["xtrabackup_version"][:3] == self.state["backup_xtrabackup_version"][:3]  # pylint: disable=E1136
+    def xtrabackup_version_match(self) -> bool:
+        installed_xtrabackup_version = get_xtrabackup_version()
+        backup_xtrabackup_version = self.state["backup_xtrabackup_version"]
+        if not backup_xtrabackup_version:
+            return True
+        return installed_xtrabackup_version[:3] == backup_xtrabackup_version[:3]
```

myhoard/basebackup_restore_operation.py
```diff
@@ -56,11 +56,15 @@ def __init__(
...
 
     def restore_backup(self):
         if os.path.exists(self.mysql_data_directory):
             raise ValueError(f"MySQL data directory {self.mysql_data_directory!r} already exists")
 
+        self.xtrabackup_version = get_xtrabackup_version()
+        self.log.info("XtraBackup version: %r", self.xtrabackup_version)
```
ettanany, line 67: As mentioned above, I do not think this is needed here.

test/test_util.py
```diff
@@ -385,4 +385,24 @@ def test_restart_unexpected_dead_sql_thread() -> None:
 
 def test_trabackup_version() -> None:
```
ettanany, line 386: Would be nice to fix this typo :)
```suggestion
def test_xtrabackup_version() -> None:
```

myhoard/controller.py
```diff
@@ -1224,7 +1224,11 @@ def _handle_mode_restore(self):
         self._process_local_binlog_updates()
         self._extend_binlog_stream_list()
         if self.restore_coordinator.phase is RestoreCoordinator.Phase.failed_basebackup:
-            self._mark_failed_restore_backup_as_broken()
+            # If a backup fails to restore, but the Percona PXB version used to create the backup
+            # differs from the current version, we do not mark the backup as "broken"
+            # The issue may be related to Percona PXB version compatibility.
+            if self.restore_coordinator.xtrabackup_version_match in (True, None):
```
ettanany, line 1230:
```suggestion
            if self.restore_coordinator.xtrabackup_version_match:
```

myhoard/restore_coordinator.py
```diff
@@ -315,6 +319,12 @@ def is_complete(self) -> bool:
     def phase(self) -> "RestoreCoordinator.Phase":
         return self.state["phase"]
 
+    @property
+    def xtrabackup_version_match(self) -> bool | None:
+        if self.state["xtrabackup_version"] is None or self.state["backup_xtrabackup_version"] is None:
+            return None
+        return self.state["xtrabackup_version"][:3] == self.state["backup_xtrabackup_version"][:3]  # pylint: disable=E1136
```
ettanany, line 326:
```suggestion
    def xtrabackup_version_match(self) -> bool:
        installed_xtrabackup_version = get_xtrabackup_version()
        backup_xtrabackup_version = self.state["backup_xtrabackup_version"]
        if not backup_xtrabackup_version:
            return True
        return installed_xtrabackup_version[:3] == backup_xtrabackup_version[:3]
```

ettanany (dismissed):

myhoard/controller.py
egor-voynov-aiven: We have 3 cases:
1. When versions match - `xtrabackup_version_match == True`
2. When versions don't match - `xtrabackup_version_match  == False`
3. When we can't extract version of the backup -  `xtrabackup_version_match  == None`

We don't want mark a backup as broken only in the second case.
For your suggestion we won't mark backup as broken in 3 case.

myhoard/restore_coordinator.py
egor-voynov-aiven: If we can't extract `backup_xtrabackup_version` we shouldn't return `True`. It could be more confusing than return None in the case.

ettanany: It depends on how you see it. If `installed_xtrabackup_version` is `x.y.z` and `backup_xtrabackup_version` is `None`, to me that still means they do not match.

ettanany: But that's really not an issue, if you do not like to limit the return type to `bool`, I think we can just focus on whether we need to store `installed_xtrabackup_version` in state, unless I missed something we do not need that.

PR #200, opened 2025-01-23
Apply some dependency cleanup

ettanany:

Clean up a bit our dev and runtime pip dependencies.


egor-voynov-aiven (approved):
:+1:

PR #201, opened 2025-02-03
Add support for incremental backups [BF-3005]

alexole:
* Add support for incremental backup (via backup endpoint and/or schedule). This helps reduce costs of object storage and potentially speed up the backup (however the restoration will take longer as more than one backup needs to be restored).

* Disable mark as broken automatic functionality

* Add incremental tag to backup/restoration metrics

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ backup daemon for PostgreSQL.
   standby servers)
 - Almost no extra local disk space requirements for creating and restoring
   backups
+- Incremental backups
 
 Fault-resilience and monitoring:
 
@@ -214,6 +215,18 @@ Name of the backup site to which new backups should be created to. See
 `backup_sites` for more information. Only needs to be defined if multiple
 non-recovery backup sites are present.
 
+**incremental.enabled**
+
+Boolean setting, which controls periodic incremental backups, according to the schedule
+`incremental.full_backup_week_schedule`
+
+**incremental.full_backup_week_schedule**
+
+A string of comma-separated days of the week: `mon`, `tue`, `wed`, `thu`, `fri`, `sat`, `sun`.
+Defines on which days full backups should be taken, thus other days incremental backups will be done.
+E.g. if the value of this setting is `sun,wed`, then full backup is taken on Sundays and Wednesdays, other days of the
+week incremental backups will be scheduled. Requires `"incremental.enabled": true`.
+
 **backup_sites**
 
 Object storage configurations and encryption keys. This is an object with
@@ -343,6 +356,11 @@ here.
 The full path and file name prefix of relay logs. This must be the same as the
 corresponding MySQL configuration option except full path is always required
 here.
+
+**restore_auto_mark_backups_broken**
+
+Boolean value. Enables a mechanism to automatically mark backups as broken if restoration fails for
+multiple times in a row. Defaults to `false`
 
 **restore_free_memory_percentage**
 
@@ -526,7 +544,8 @@ binary log file. Request body must be like this:
 ```
 {
   "backup_type": "{basebackup|binlog}",
-  "wait_for_upload": 3.0
+  "wait_for_upload": 3.0,
+  "incremental": false
 }
 ```
 
@@ -543,6 +562,12 @@ backed up.
 This is only valid in case `backup_type` is `binlog`. In that case the
 operation will block for up to as many seconds as specified by this parameter
 for the binary log upload to complete before returning.
+
+**incremental**
+
+Boolean value, when `true` - requests incremental base backup if possible.
+Setting is used only in combination with `basebackup` backup_type.
+Defaults to `false`.
 
 Response on success looks like this:
 
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -79,6 +79,11 @@ class BaseBackupFailureReason(str, enum.Enum):
     xtrabackup_error = "xtrabackup_error"
 
 
+class IncrementalBackupInfo(TypedDict):
+    last_checkpoint: str | None
+    required_streams: List[str] | None
+
+
 class BackupStream(threading.Thread):
     """Handles creating a single consistent backup stream. 'stream' here refers to uninterrupted sequence
     of backup data that can be used to restore the system to a consistent state. It includes the basebackup,
@@ -132,6 +137,7 @@ class BackupStream(threading.Thread):
         backup_reason: Optional["BackupStream.BackupReason"]
         created_at: float
         immediate_scan_required: bool
+        incremental: bool
         initial_latest_complete_binlog_index: Optional[int]
         last_binlog_upload_time: int
         last_processed_local_index: Optional[int]
@@ -181,6 +187,7 @@ class BackupStream(threading.Thread):
         temp_dir: str,
         xtrabackup_settings: Optional[Dict[str, int]] = None,
         split_size: Optional[int] = 0,
+        incremental_backup_info: IncrementalBackupInfo | None = None,
     ) -> None:
         super().__init__()
         stream_id = stream_id or self.new_stream_id()
@@ -195,6 +202,7 @@ class BackupStream(threading.Thread):
         self.file_storage_setup_fn = file_storage_setup_fn
         self.file_storage: Optional["BaseTransfer"] = None
         self.file_uploaded_callback = file_uploaded_callback
+        self.incremental_backup_info = incremental_backup_info
         self.is_running = True
         self.iteration_sleep = BackupStream.ITERATION_SLEEP
         self.last_basebackup_attempt: Optional[float] = None
@@ -236,6 +244,7 @@ class BackupStream(threading.Thread):
             "backup_reason": backup_reason,
             "created_at": time.time(),
             "immediate_scan_required": False,
+            "incremental": incremental_backup_info is not None,
             "initial_latest_complete_binlog_index": latest_complete_binlog_index,
             "last_binlog_upload_time": 0,
             "last_processed_local_index": None,
@@ -674,7 +683,9 @@ class BackupStream(threading.Thread):
             # a separate measurement because uploads are not ongoing all the time and calculating
             # rate based on raw byte counter requires knowing when the operation started and ended
             self.stats.gauge_int(
-                "myhoard.backup_stream.basebackup_bytes_uploaded", self.basebackup_bytes_uploaded + bytes_sent
+                "myhoard.backup_stream.basebackup_bytes_uploaded",
+                self.basebackup_bytes_uploaded + bytes_sent,
+                tags={"incremental": self.is_incremental},
             )
             last_value[0], last_time[0] = track_rate(
                 current=bytes_sent,
@@ -707,6 +718,12 @@ class BackupStream(threading.Thread):
             # changes in rohmu.
             uploaded_size = file_storage.get_file_size(storage_file_name)
             self.basebackup_bytes_uploaded += uploaded_size
+
+            self.stats.gauge_int(
+                "myhoard.backup_stream.basebackup_bytes_uploaded",
+                self.basebackup_bytes_uploaded,
+                tags={"incremental": self.is_incremental},
+            )
 
             # don't loop forever if we can't upload anything
             if not self.split_size or self.split_size < 1:
@@ -980,6 +997,10 @@ class BackupStream(threading.Thread):
     def _should_list_with_metadata(self, *, next_index: int) -> bool:
         return next_index % BINLOG_BUCKET_SIZE == 0 or next_index == 1 or not isinstance(self.file_storage, S3Transfer)
 
+    @property
+    def is_incremental(self) -> bool:
+        return self.incremental_backup_info is not None
+
     def _take_basebackup(self) -> None:
         assert self.file_storage is not None
         if self.last_basebackup_attempt is not None:
@@ -1014,6 +1035,9 @@ class BackupStream(threading.Thread):
             stats=self.stats,
             stream_handler=self._basebackup_stream_handler,
             temp_dir=self.temp_dir,
+            incremental_since_checkpoint=(
+                self.incremental_backup_info.get("last_checkpoint") if self.incremental_backup_info else None
+            ),
         )
         try:
             self.basebackup_operation.create_backup()
@@ -1070,12 +1094,17 @@ class BackupStream(threading.Thread):
                 "binlog_name": binlog_info["file_name"] if binlog_info else None,
                 "binlog_position": binlog_info["file_position"] if binlog_info else None,
                 "backup_reason": self.state["backup_reason"],
+                "checkpoints_file_content": self.basebackup_operation.checkpoints_file_content,
                 "compressed_size": compressed_size,
                 "encryption_key": rsa_encrypt_bytes(self.rsa_public_key_pem, encryption_key).hex(),
                 "end_size": self.basebackup_operation.data_directory_size_end,
                 "end_ts": end_time,
                 "gtid": last_gtid,
                 "gtid_executed": gtid_executed,
+                "incremental": self.basebackup_operation.incremental_since_checkpoint is not None,
+                "required_streams": (
+                    self.incremental_backup_info.get("required_streams") if self.incremental_backup_info else None
+                ),
                 "initiated_at": self.created_at,
                 "lsn_info": self.basebackup_operation.lsn_info,
                 "normalized_backup_time": self.state["normalized_backup_time"],
@@ -1084,6 +1113,7 @@ class BackupStream(threading.Thread):
                 "number_of_splits": basebackup_file_metadata.get("number_of_splits", 1),
                 "start_size": self.basebackup_operation.data_directory_size_start,
                 "start_ts": start_time,
+                "tool_version": self.basebackup_operation.tool_version,
                 "uploaded_from": self.server_id,
             }
             self.file_storage.store_file_from_memory(
--- a/myhoard/basebackup_restore_operation.py
+++ b/myhoard/basebackup_restore_operation.py
@@ -1,9 +1,14 @@
 # Copyright (c) 2019 Aiven, Helsinki, Finland. https://aiven.io/
 from .errors import DiskFullError
-from .util import BinVersion, find_extra_xtrabackup_executables, get_xtrabackup_version, parse_version, parse_xtrabackup_info
+from .util import (
+    CHECKPOINT_FILENAME,
+    find_extra_xtrabackup_executables,
+    get_xtrabackup_version,
+    parse_version,
+)
 from contextlib import suppress
 from rohmu.util import increase_pipe_capacity, set_stream_nonblocking
-from typing import Final, Optional
+from typing import Dict, Final, Optional
 
 import base64
 import fnmatch
@@ -38,7 +43,9 @@ class BasebackupRestoreOperation:
         mysql_data_directory,
         stats,
         stream_handler,
-        temp_dir,
+        target_dir: str,
+        temp_dir: str,
+        backup_tool_version: str | None = None,
     ):
         self.current_file = None
         self.data_directory_size_end = None
@@ -51,34 +58,37 @@ class BasebackupRestoreOperation:
         self.mysql_data_directory = mysql_data_directory
         self.number_of_files = 0
         self.prepared_lsn = None
-        self.proc = None
+        self.proc: subprocess.Popen[bytes] | None = None
         self.stats = stats
         self.stream_handler = stream_handler
-        self.temp_dir = None
-        self.temp_dir_base = temp_dir
-        self.backup_xtrabackup_info = None
+        self.target_dir = target_dir
+        self.temp_dir = temp_dir
+        self.backup_xtrabackup_info: Dict[str, str] | None = None
+        self.backup_tool_version = backup_tool_version
 
-    def restore_backup(self):
-        if os.path.exists(self.mysql_data_directory):
-            raise ValueError(f"MySQL data directory {self.mysql_data_directory!r} already exists")
-
+    def prepare_backup(
+        self, incremental: bool = False, apply_log_only: bool = False, checkpoints_file_content: str | None = None
+    ):
         # Write encryption key to file to avoid having it on command line. NamedTemporaryFile has mode 0600
-        with tempfile.NamedTemporaryFile() as encryption_key_file:
-            encryption_key_file.write(base64.b64encode(self.encryption_key))
-            encryption_key_file.flush()
 
-            cpu_count = multiprocessing.cpu_count()
+        incremental_dir = None
+        try:
+            if incremental:
+                incremental_dir = tempfile.mkdtemp(dir=self.temp_dir, prefix="myhoard_inc_")
 
-            self.temp_dir = tempfile.mkdtemp(dir=self.temp_dir_base)
+            with tempfile.NamedTemporaryFile() as encryption_key_file:
+                encryption_key_file.write(base64.b64encode(self.encryption_key))
+                encryption_key_file.flush()
 
-            try:
+                cpu_count = multiprocessing.cpu_count()
+                extract_dir = self.target_dir if not incremental_dir else incremental_dir
                 command_line = [
                     "xbstream",
                     # TODO: Check if it made sense to restore directly to MySQL data directory so that
                     # we could skip the move phase. It's not clear if move does anything worthwhile
                     # except skips a few extra files, which could instead be deleted explicitly
                     "--directory",
-                    self.temp_dir,
+                    extract_dir,
                     "--extract",
                     "--decompress",
                     "--decompress-threads",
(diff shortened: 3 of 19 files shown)


myhoard/controller.py
```diff
@@ -1273,13 +1338,24 @@ def _mark_periodic_backup_requested_if_interval_exceeded(self):
             )
             and (not most_recent_scheduled or time.time() - most_recent_scheduled >= half_backup_interval_s)
         ):
+            incremental_backup_info = None
+            if self._should_schedule_incremental_backup():
+                incremental_backup_info = self.get_incremental_backup_info()
+                if not incremental_backup_info:
+                    self.log.warning(
+                        "Incremental backup is configured but not possible to take, proceeding with full backup"
+                    )
```
packi, line 1347: This should probably emit a metric? Also the decision if we're doing an incremental or full-backup farther up.

packi (commented):
Initial review pass.

myhoard/restore_coordinator.py
```diff
@@ -409,61 +415,141 @@ def get_backup_info(self) -> None:
         basebackup_info = self._load_file_data("basebackup.json")
         if not basebackup_info:
             return
+
+        required_backups = []
+        if basebackup_info["incremental"]:
+            # We need to load all backups infos until the latest full one
+            for stream_id in basebackup_info["required_streams"]:
+                info = self._load_file_data("basebackup.json", stream_id=stream_id)
```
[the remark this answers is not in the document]
alexole: I am not sure I understood that. I am not changing this part, there were no extra checks before and it's actually not needed. If myhoard is requested to restore this backup, at this point it's not very relevant to check if it's good or not.

myhoard/basebackup_restore_operation.py
```diff
@@ -51,34 +52,36 @@ def __init__(
...
 
-            try:
                 command_line = [
                     "xbstream",
                     # TODO: Check if it made sense to restore directly to MySQL data directory so that
                     # we could skip the move phase. It's not clear if move does anything worthwhile
                     # except skips a few extra files, which could instead be deleted explicitly
                     "--directory",
-                    self.temp_dir,
+                    self.target_dir if not incremental_dir else incremental_dir,
```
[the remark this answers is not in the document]
alexole: Yes, it does.

myhoard/controller.py
```diff
@@ -864,6 +877,7 @@ def _create_restore_coordinator_if_missing(self):
         backup_site = self._lookup_backup_site(options["site"])
         storage_config = backup_site["object_storage"]
         self.log.info("Creating new restore coordinator")
+        # TODO site change triggers full backup
```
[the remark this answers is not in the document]
alexole: Still relevant, nice to have feature. Not yet decided if I should do it in this scope.

myhoard/util.py
```diff
@@ -39,6 +39,9 @@
     "register_redo_log_consumer": False,
 }
 
+DOW_TO_IDX = {"mon": 0, "tue": 1, "wed": 2, "thu": 3, "fri": 4, "sat": 5, "sun": 6}
```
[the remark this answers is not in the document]
alexole: Yeah, that's probably more beautiful. Initially I wasn't sure about format and wanted to have explicit visibility on index, but I might change to what you suggested.

myhoard/controller.py
alexole: Probably, I am thinking if this is any useful information. In my mind, failing back to full backup is a normal scenario.
I am wondering how this metric could be used?

packi: Well, it would point to an issue. I think the main use would be to annotage all metrics with a tag if it's a full one or an incremental one so we can display that somewhere and can run statistics on that.

README.md
```diff
@@ -214,6 +215,18 @@ Name of the backup site to which new backups should be created to. See
...
+
+Boolean setting, which controls periodic incremental backups, according to the schedule
+`incremental.full_backup_week_schedule`
+
+**incremental.full_backup_week_schedule**
+
+A string of comma-separated days of the week: `mon`, `tue`, `wed`, `thu`, `fri`, `sat`, `sun`.
+Defines on which days full backups should be taken, thus other days incremental backups will be done.
+E.g. if the value of this setting is `sun,wed`, then full backup is taken on Sundays and Wednesdays, other days of the
+week incremental backups will be scheduled. Requires `"incremental.enabled": true`.
```
egor-voynov-aiven, line 228: Might we want to have basebackup less often than once a week?

alexole: I don't think it makes sense in practice, as it will cost restoration time.

myhoard/controller.py
```diff
@@ -1248,6 +1265,54 @@ def _mark_failed_restore_backup_as_broken(self) -> None:
 
         self._build_backup_stream(broken_backup).mark_as_broken()
 
+    def _should_schedule_incremental_backup(self) -> bool:
+        incremental_settings = self.backup_settings.get("incremental", {})
+        if not incremental_settings.get("enabled", False):
+            self.log.info("Incremental backup is disabled in configuration")
+            return False
+
+        dow_idx = datetime.datetime.now(datetime.timezone.utc).weekday()
```
dmitry-potepalov, line 1274: I guess the `utc` part might cause unexpected behavior if the service is in e.g. Sydney (+11)... Could we maybe try for "fixed interval between full backups that's reset every week"? With a full backup either every Monday or every Sunday?..

dmitry-potepalov (commented):
I still want to look at the purge logic and the tests :)
Besides that, I do not see any hard blockers except maybe the UTC issue.
My comments are mostly nits/suggestions for future improvements.

myhoard/backup_stream.py
```diff
@@ -1070,12 +1082,17 @@ def _take_basebackup(self) -> None:
                 "binlog_name": binlog_info["file_name"] if binlog_info else None,
                 "binlog_position": binlog_info["file_position"] if binlog_info else None,
                 "backup_reason": self.state["backup_reason"],
+                "checkpoints_file_content": self.basebackup_operation.checkpoints_file_content,
```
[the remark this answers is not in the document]
alexole: I was thinking about that, but it's nice for investigation/debugging purposes to see that in the state file. It's not a binary file, rather a text in a key=value format with some useful values.

myhoard/controller.py
```diff
@@ -1248,6 +1265,54 @@ def _mark_failed_restore_backup_as_broken(self) -> None:
...
+        self.log.info("According to `full_backup_week_schedule` full backup should be scheduled (day: %r)", dow_idx)
+        return False
+
+    def get_incremental_backup_info(self) -> IncrementalBackupInfo | None:
+        """
+        Incremental backup is possible when all the backups starting from the recent full backup have checkpoints stored
+        """
+        self.log.warning("Determining if incremental backup is possible")
+
+        backups = sort_completed_backups(self.state["backups"], reverse=True)
```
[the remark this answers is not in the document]
alexole: The backups are re-fetched after node replacement and put into state, so we have a proper backup list as before and therefore incremental backup will be possible.

myhoard/backup_stream.py
```diff
@@ -132,6 +137,7 @@ class State(TypedDict):
         backup_reason: Optional["BackupStream.BackupReason"]
         created_at: float
         immediate_scan_required: bool
+        incremental: bool
```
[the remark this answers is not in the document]
alexole: Maybe it should be names differently, but it's actually used mainly for understanding if incremental backup was requested for debugging purposes whatnot. It will be saved in the backup stream state.
https://github.com/Aiven-Open/myhoard/pull/201/files#diff-aa4e1b9f013727fc74f02d5af72a9eb70b52055958f73173e002e6711cb41ea3R1092

myhoard/controller.py
```diff
@@ -1248,6 +1265,54 @@ def _mark_failed_restore_backup_as_broken(self) -> None:
...
+        required_streams: List[str] = []
+        prev_backup = None
+        for backup in backups:
+            required_streams.append(backup["stream_id"])
```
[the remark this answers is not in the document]
alexole: I don't think skipping broken incremental backup is something easily doable, but I see your point. I can move this line right before, not sure why I put it there:
```
            if not backup["basebackup_info"].get("incremental"):
```

myhoard/controller.py
```diff
@@ -1248,6 +1265,54 @@ def _mark_failed_restore_backup_as_broken(self) -> None:
...
+                return None
+            if not backup["basebackup_info"].get("checkpoints_file_content"):
+                self.log.warning(
+                    "Incremental backup is not possible - backup %r does not have a checkpoint", backup["stream_id"]
+                )
+                return None
+            if not backup["basebackup_info"].get("incremental"):
+                self.log.info("Incremental backup requirements are met.")
+                return {
+                    "last_checkpoint": prev_backup["basebackup_info"]["checkpoints_file_content"],
```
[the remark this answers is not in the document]
alexole: `prev_backup` is only filled once, it's basically the backup which we should create the current incremental backup based on. `backup` is just a loop variable to go through the backups until/if we meet full backup or any other quit condition.

README.md
```diff
@@ -214,6 +215,18 @@ Name of the backup site to which new backups should be created to. See
 `backup_sites` for more information. Only needs to be defined if multiple
 non-recovery backup sites are present.
 
+**incremental.enabled**
+
+Boolean setting, which controls periodic incremental backups, according to the schedule
```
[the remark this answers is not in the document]
alexole: It mentions dependency on `incremental.full_backup_week_schedule` which explains this part, let me know if that's enough.
https://github.com/Aiven-Open/myhoard/pull/201/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R226

myhoard/controller.py
```diff
@@ -1248,6 +1265,54 @@ def _mark_failed_restore_backup_as_broken(self) -> None:
...
+
+        dow_idx = datetime.datetime.now(datetime.timezone.utc).weekday()
+        dow_schedule = incremental_settings.get("full_backup_week_schedule")
+        if dow_idx not in parse_dow_schedule(dow_schedule):
+            self.log.info(
+                "According to `full_backup_week_schedule` incremental backup should be scheduled (day: %r)", dow_idx
+            )
+            return True
+
+        self.log.info("According to `full_backup_week_schedule` full backup should be scheduled (day: %r)", dow_idx)
```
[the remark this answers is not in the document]
alexole: There is a return statement after the previous log. And those messages are different. Or what do you mean?

myhoard/restore_coordinator.py
```diff
@@ -410,61 +416,141 @@ def get_backup_info(self) -> None:
...
+        start_time = time.monotonic()
+        restore_basebackup_info = self.state["basebackup_info"]
+        required_backups = self.state.get("required_backups", [])
+
+        with tempfile.TemporaryDirectory(dir=self.temp_dir, prefix="myhoard_target_") as temp_target_dir:
+            try:
+                if required_backups:
+                    self.log.info(
+                        "Restoring previous basebackups required by currently restored backup: %r",
+                        restore_basebackup_info["required_streams"],
```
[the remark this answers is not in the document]
alexole: sorry, I am not sure what you mean here.

myhoard/controller.py
alexole: I see the point, the fact is that all the date time configuration is in UTC already, so backup hour/minute is assumed to be UTC. Thus I think it's OK to assume this is UTC also. If we are to change the way we deal with time, then we will need to change this together.

myhoard/backup_stream.py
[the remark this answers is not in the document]
dmitry-potepalov: ah, I thought it was a member of BackupStream for some reason, but it's actually in State, nvm

myhoard/backup_stream.py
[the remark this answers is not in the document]
dmitry-potepalov: > It's not a binary file

... for now :) I just don't want to introduce a context for an _interesting_ bug in the future. But I don't insist, your call.

myhoard/controller.py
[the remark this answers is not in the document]
dmitry-potepalov: > those messages are different.

Indeed, only a single word... Maybe just omit the "According to full_backup_week_schedule" part entirely and just leave Full backup/Incremental backup scheduled?..

myhoard/controller.py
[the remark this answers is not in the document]
dmitry-potepalov: I mean - if the backup is broken, it's definitely not required -> why add a broken backup to the `required_streams` list? Maybe just move the `if backup.get("broken_at")` to the start of the iteration:
```
 for backup in backups:
            if backup.get("broken_at"):
                self.log.warning("Incremental backup is not possible - found broken backup %r", backup["stream_id"])
                return None
...
```

myhoard/restore_coordinator.py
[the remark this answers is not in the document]
dmitry-potepalov: I meant that it looks a bit confusing that we log `restore_basebackup_info["required_streams"]`, but are checking and using `required_backups`; what if the logic for picking `required_backups` changes somehow and yields a smaller list/different list? (looks unlikely from where we are now, but in general I think that logging the thing in use/some list of ids derived from it is a safer approach)

myhoard/controller.py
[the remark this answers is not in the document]
alexole: I see, in this case no need to add if the checkpoint is missing too, that's why I am suggesting to move it to place mentioned above :)

myhoard/controller.py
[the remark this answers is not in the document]
alexole: Maybe it's more idiomatic, but I am iterating through the backups anyway, so why not just to add the `if` :)

myhoard/util.py
```diff
@@ -40,6 +40,10 @@
     "register_redo_log_consumer": False,
 }
 
+# Indexes of this array are used for mapping the days (0..6), order is important
+DOW_ORDERED = ["mon", "tue", "wed", "thu", "fri", "sat", "sun"]
```
ettanany, line 44: nit: I think `dow_to_idx` dict used in the other repo reads better as it's self-explanatory without a comment :)

alexole: hehe, I can't make everyone happy :) I did use dict initially and then changed according to 
https://github.com/Aiven-Open/myhoard/pull/201#discussion_r1961549411

myhoard/controller.py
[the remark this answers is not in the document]
alexole: Addressed.

myhoard/restore_coordinator.py
```diff
@@ -410,61 +416,141 @@ def get_backup_info(self) -> None:
...
+                        stream_id, required_basebackup_info = required_backup
+                        current_start_time = time.monotonic()
+                        self.log.info(
+                            "Preparing required backup (incremental: %r): %r (%r/%r)",
+                            required_basebackup_info["incremental"],
+                            stream_id,
+                            idx,
+                            len(required_backups),
+                        )
+                        assert required_basebackup_info["incremental"] == (idx != 1)
```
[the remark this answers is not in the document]
alexole: I've moved this validation into getting_backup_info phase. That should be sufficient.

README.md
alexole: Closing it unless you want to discuss further.

myhoard/controller.py
```diff
@@ -1752,6 +1910,7 @@ def _start_new_backup(self, *, backup_reason: BackupStream.BackupReason, normali
             temp_dir=self.temp_dir,
             xtrabackup_settings=self.xtrabackup_settings,
             split_size=backup_site.get("split_size", 0),
+            incremental_backup_info=incremental_backup_info,
```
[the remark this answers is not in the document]
alexole: Should not be the case.

myhoard/restore_coordinator.py
```diff
@@ -410,61 +416,141 @@ def get_backup_info(self) -> None:
...
+                        self._run_basebackup_restore_operation(
+                            required_basebackup_info,
+                            target_dir=temp_target_dir,
+                            temp_dir=self.temp_dir,
+                            stream_id=stream_id,
+                            prepare_only=True,
+                            apply_log_only=True,
+                            incremental=(idx != 1),  # Only the first one should be full
+                        )
+                        self.update_state(required_backups_restored=idx)
```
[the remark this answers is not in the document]
alexole: This is not used for now, just for info purposes.
Re-downloading/preparing is a bit wasteful but the thing is applying the incremental backup is not idempotent, if it failed while applying, retrying will most certainly not work and we are in an inconsistent state, we need to start over unfortunately. I agree we could try catch downloading errors only and re-start download for this specific backup, but I would keep it out of scope for this ticket at least. Let's do it as an improvement after this main work is merged.

myhoard/controller.py
alexole: That's a good idea about the tags and I added that to metrics related to backup/restoration. I still think this metric is a bit redundant, no sure how can I use it and it will be noisy anyway as in the beginning without some metadata only full backups will be take, so the fallback will take place.
Let me know if you are OK with change/answer.

alexole: I am resolving it for now unless you have a strong opinion :)

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
 
-        if non_broken_backups_count <= self.backup_settings["backup_count_max"] < len(purgeable):
+        # If the first backup is incremental, it's useless without a preceding full backup, so drop it
+        if is_backup_incremental(backup):
+            return backup
+
+        # Do not consider broken backups for the count, they will still be purged,
+        # but we should only purge when the count of non-broken backups has exceeded the limit.
+        non_broken_backups_count = Controller.count_non_broken_backups(completed_backups)
+        if non_broken_backups_count <= self.backup_settings["backup_count_max"] < len(completed_backups):
```
kathia-barahona, line 1460: **question** Why not consider only full backups on the `backup_count_max`?  I mean incremental backups are just small bits of another backup. Also, the log info message might be a bit misleading as its not clear how many of the backups are full or incremental.

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
+    @enum.unique
+    class NoPurgeReason(str, enum.Enum):
+        no_completed_backups = "no_completed_backups"
+        within_max_amount = "within_max_amount"
+        min_amount_required = "min_amount_required"
+        non_closed_backup = "non_closed_backup"
+        preserved_backup = "preserved_backup"
+        required_backup_within_retention = "required_backup_within_retention"
+        requirements_not_met = "requirements_not_met"
+        active_backup = "active_backup"
```
kathia-barahona, line 1436: **nit-pick** Could you move this to the start of the module? I think the enum names are quite self-explanatory, so I don't find it useful to have it close to the purging functions

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
+
+    def _get_backup_to_purge(self) -> Backup | NoPurgeReason:
```
kathia-barahona, line 1438: 🙈 Not a fan of that return type.

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
+            return Controller.NoPurgeReason.within_max_amount
 
-        if non_broken_backups_count <= self.backup_settings["backup_count_min"]:
-            return
+        # Create a list of incremental and dependant backups, they are interconnected and needed for later checks
+        incremental_backups = []
+        for b in completed_backups[1:]:
+            if not is_backup_incremental(b):
+                break
+            incremental_backups.append(b)
```
kathia-barahona, line 1475: **suggestion** How about having a function returning completed backups grouped by full backup plus incremental. Something like:

```
[
   {
      "full_backup": Backup(....),
      "incrementals": [Backup(...), ...],
   },
   {
     "full_backup": Backup(...),
     "incrementals": []
   },
]
```
This way we already know which are the incremental for the oldest backup, also it makes it a bit easier for `count_non_broken_backups` as well

kathia-barahona: In case for "orphaned" incremental, it must happen when there's no "full_backup" and we should just purge all together

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
+            return b["basebackup_info"].get("incremental", False)
 
-        if non_broken_backups_count <= self.backup_settings["backup_count_max"] < len(purgeable):
+        # If the first backup is incremental, it's useless without a preceding full backup, so drop it
+        if is_backup_incremental(backup):
+            return backup
+
+        # Do not consider broken backups for the count, they will still be purged,
+        # but we should only purge when the count of non-broken backups has exceeded the limit.
+        non_broken_backups_count = Controller.count_non_broken_backups(completed_backups)
```
kathia-barahona, line 1459: **nit-pick** (can just ignore)

```python
non_broken_backups_count = self.count_non_broken_backups(completed_backups)
```

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
-        purgeable = sort_completed_backups(purgeable)
-        backup = purgeable[0]
+        dependant_backups = [backup] + incremental_backups
+        # We should always keep min amount of backups, check if that works if we drop this and potentially following
+        # incremental backups
+        if (
+            non_broken_backups_count - Controller.count_non_broken_backups(dependant_backups)
+            < self.backup_settings["backup_count_min"]
+        ):
+            return Controller.NoPurgeReason.min_amount_required
```
kathia-barahona, line 1484: Same here, I think `backup_count_min` should mean count of full backups.

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
+            with self.lock:
+                self.state_manager.update_state(stream_to_be_purged=None)
+                current_backups = [
+                    current_backup
+                    for current_backup in self.state["backups"]
+                    if current_backup["stream_id"] != backup["stream_id"]
+                ]
+                self.state_manager.update_state(backups=current_backups)
+                owned_stream_ids = [sid for sid in self.state["owned_stream_ids"] if sid != backup["stream_id"]]
+                self.state_manager.update_state(owned_stream_ids=owned_stream_ids)
```
kathia-barahona, line 1548: **comment** Is it possible to just purge everything at once? I mean full backup + dependents?

kathia-barahona (commented):
Haven't reviewed everything. But have some comments related to the changes in the Controller. I'll later add reviews related to restoration and backup creation

myhoard/controller.py
```diff
@@ -1310,67 +1409,143 @@ def _process_removed_binlogs(self, binlogs):
...
+        # to drop they will be dropped soon enough
+        backup = completed_backups[0]
+
+        def is_backup_incremental(b: Backup) -> bool:
+            return b["basebackup_info"].get("incremental", False)
 
-        if non_broken_backups_count <= self.backup_settings["backup_count_max"] < len(purgeable):
+        # If the first backup is incremental, it's useless without a preceding full backup, so drop it
+        if is_backup_incremental(backup):
+            return backup
```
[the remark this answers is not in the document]
kathia-barahona: You can dismiss this, I'd just noticed we only purge a backup at once. I left a comment related to that as well

myhoard/controller.py
alexole: Incremental backup which is valid and has requiring backups should also count as in the logics before.
If we change all the data constantly it's not bits of another backup, in theory it could equal the full backup.

myhoard/controller.py
alexole: Same answer I guess :)

myhoard/controller.py
alexole: Not ideal I guess, but could not come with with a better idea.

myhoard/controller.py
alexole: Well, maybe, I need to have a look, I did not want to change this functionality as we did drop only one backup at a time before.

myhoard/controller.py
kathia-barahona: I mean, knowing the full backup is broken, inherently makes the dependents broken. An  incremental backup can also be marked as broken?

myhoard/controller.py
kathia-barahona: yes, but still a single incremental backup doesn't make sense without a full backup, I mean you can't restore from an incremental. So, for example having `backup_count_max=5` no longer makes sense. As if we only take a full backup once per week, then we will easily exceed this amount on the 6th day.

kathia-barahona: I think we might need a validation for `backup_count_max`, as setting too low might bring issues, right? But also, it can be dangerous if full backup days are switched

alexole: There will be no hanging incremental backup, by this line it will be purged.

If we have max backups 5, one full backup and 6 incremental ones, this if statement will be false and we will not drop it because exactly if we do - incremental will be invalid.
I am sorry but I can't understand where the problem with the logic is? :) Could you tell exactly with which parameters behaviour will be wrong?

alexole: But isn't it how it should be? We leave 2 - minimal amount of backups, so service has 2 (not 1) restorable backups. They are two valid backups with the exception that the 7th depends on the 6th. I don't see the problem here. Treat an incremental backup as just a normal base backup with a dependency.

kathia-barahona: Well, if that's expected then okay :)

dmitry-potepalov (approved):
I cannot say that this looks good to me, I think that this needs a "complexity reduction" pass, some of the old logic that was sort-of-OK without incremental backups now becomes even messier. But I don't think it's a valid enough reason to block merging it.

PR #203, opened 2025-02-05
Clean up pre-commit checks and lint pipeline

egor-voynov-aiven:


myhoard/binary_io_slice.py
```diff
@@ -74,10 +74,10 @@ def truncate(self, __size: int | None = None) -> int:
     def writable(self) -> bool:
         return False
 
-    def write(self, __s: bytes) -> int:
+    def write(self, __s: bytes) -> int:  # type: ignore
```
ettanany, line 77: I think it would be nice to be explicit about what is ignored exactly (same applies to other places where we also have this).

egor-voynov-aiven: Error messages are too long, I can specify particular error type like: ` # type: ignore[override]`

ettanany (approved):

PR #204, opened 2025-02-13
Use the same PXB version for restoring the backup as was used for its creation

egor-voynov-aiven:
Added a new environment variable `PXB_EXTRA_BIN_PATHS`, which allows specifying additional paths to PXB binaries. These binaries can be used for restoring backups with the corresponding version.

### DIFF
--- a/myhoard/util.py
+++ b/myhoard/util.py
@@ -5,9 +5,10 @@ from cryptography.hazmat.primitives.asymmetric import padding, rsa
 from cryptography.hazmat.primitives.hashes import SHA1
 from logging import Logger
 from math import log10
+from pathlib import Path
 from pymysql.connections import Connection
 from pymysql.cursors import DictCursor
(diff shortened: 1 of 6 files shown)


myhoard/util.py
```diff
@@ -683,3 +690,18 @@ def file_name_for_basebackup_split(base_file_name: str, split_nr: int) -> str:
...
+
+
+def find_extra_xtrabackup_executables() -> list[BinInfo]:
+    raw_paths = os.environ.get("PXB_EXTRA_BIN_PATHS")
+    if not raw_paths:
+        return []
+    result = []
+    for extra_raw_path in raw_paths.split(os.pathsep):
+        extra_path = Path(extra_raw_path)
+        if extra_path.is_dir():
```
alexole, line 702: Can't we assume it's always directory as env name suggests? I find it non intuitive if binary paths are provided and in that case we also need to add validation that the binary actually is xtrabackup.

alexole (approved):
LGTM

PR #225, opened 2026-03-19
Optimize binlog_upload_delay function

alexole:
This function used to call stat on every single pending binlog just to return an upload delay, which is wasteful and in case of huge pile of binlogs it's a significant performance dip. Instead we find the first non purged binlog and use stat of this file, as it's guaranteed to be the oldest one.


danielblasina-tech (approved):
LGTM

PR #227, opened 2026-04-22
Fix relay log position correction for backup restoration

alexole:
`_parse_gtid_executed_ranges()` uses `read_until_position=binlog_position` when scanning the binlog file for GTID events.
If `xtrabackup` reports a position that's mid-event (due to a race in performance_schema.log_status — MySQL Bug#32442772 https://github.com/mysql/mysql-server/commit/749fd742b1c83d38590afb1364fd40fd3edfcb66), the existing position-correction code can't find the next valid event boundary because reading stops too early.
This causes `CHANGE REPLICATION SOURCE TO RELAY_LOG_POS=<mid-event>` which fails with error 13121 (relay log read failure).

Remove `read_until_position` to allow scanning past the reported position, with a safety bound to prevent reading the entire file when the matching GTID isn't found.

It seems it was fixed in MySQL 8.0.42 https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-42.html

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -1348,7 +1348,7 @@ class RestoreCoordinator(threading.Thread):
         local_name = self._relay_log_name(index=binlog["adjusted_remote_index"] + self.state["binlog_name_offset"])
         gtid_infos = []
         found_last_entry = False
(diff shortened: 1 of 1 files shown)


kevinhughes27:
makes sense to me 👍

PR #228, opened 2026-04-27
Allow updating binlog purge settings while taking the backup

alexole:
When a user changes `binlog_retention_period` while a basebackup is in progress, the new setting is not applied until the backup completes (potentially hours later). This is because myhoard's SIGHUP reload does a full stop/start cycle, which is blocked by `is_safe_to_reload()` during basebackup.
This allows safe settings (`binlog_retention_period`) to be applied when reloading is not considered safe.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -262,6 +262,10 @@ class Controller(threading.Thread):
                     return False
         return True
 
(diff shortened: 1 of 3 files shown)


VitaliVinahradski (approved):
LGTM

PR #231, opened 2026-05-07
create an option to pause backups

danielblasina-tech:
create an option to pause backups

### DIFF
--- a/myhoard/web_server.py
+++ b/myhoard/web_server.py
@@ -89,6 +89,22 @@ class WebServer:
 
                 return json_response(response)
 
+    async def backup_pause(self, request):
+        with self._handle_request(name="backup_pause"):
+            body = await self._get_request_json(request)
+
+            try:
+                pause_until = datetime.fromisoformat(body.get("until"))
+            except ValueError:
+                raise BadRequest("`until` must be a valid iso-format datetime string.")
+
+            if pause_until is None or pause_until < datetime.now(timezone.utc):
+                raise BadRequest("pause_until is not a valid date")
+
+            self.controller.pause_backups(until=pause_until)
+
+            return json_response({"success": True})
+
     async def backup_preserve(self, request):
         with self._handle_request(name="backup_preserve"):
             stream_id = request.match_info["stream_id"]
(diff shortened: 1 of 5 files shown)


myhoard/web_server.py
```diff
@@ -89,6 +89,18 @@ async def backup_list(self, _request):
 
                 return json_response(response)
 
+    async def backup_pause(self, request):
+        with self._handle_request(name="backup_pause"):
+            body = await self._get_request_json(request)
+            pause_until = datetime.fromisoformat(body.get("until"))
+
+            if pause_until is None or pause_until < datetime.now(timezone.utc):
+                raise Exception("pause_until is not a valid date")
```
alexole: nit: could we use BadRequest exception? In this case it will be converted into 4xx http response rather than 5xx

myhoard/web_server.py
```diff
@@ -269,6 +281,7 @@ def _add_routes(self):
             [
                 web.get("/backup", self.backup_list),
                 web.post("/backup", self.backup_create),
+                web.put("/backup/pause", self.backup_pause),
```
alexole, line 288: Could you please update the readme too?

test/test_controller.py
```diff
@@ -1580,6 +1580,43 @@ def flush_binlogs():
...
+    # Pause backups
+    m_controller.state["pause_backups_until"] = datetime.datetime(2099, 1, 3, 3, tzinfo=datetime.timezone.utc).isoformat()
+
+    flush_binlogs()
+
+    time_machine.move_to("2023-01-05T03:00:00+00:00")
+    expected_normalized_time = datetime.datetime(2023, 1, 5, 3, tzinfo=datetime.timezone.utc)
+
+    assert m_controller._current_normalized_backup_timestamp() == expected_normalized_time.isoformat()
+    while_asserts(lambda: streaming_binlogs(m_controller, 1), timeout=10)
```
alexole, line 1617: Could you please also check that if you move time right after the pause_backups_until, it resumes backups again.

myhoard/web_server.py
```diff
@@ -269,6 +285,7 @@ def _add_routes(self):
...
```
alexole, line 288: I just thought maybe it would be also great to add an endpoint to reset the pause? (DELETE /backup/pause?) Imagine scenario when somebody made a mistake and set the wrong date, it would be great to reset it via endpoint instead of dealing with the state json. But up to you, otherwise LGTM.

danielblasina-tech: in this case one can set to date now+1 minute :)

danielblasina-tech: will add the DELETE endpoint

alexole: True.

alexole (approved):

PR #232, opened 2026-05-11
Add preparing_backup restore phase with LSN-derived progress

alexole:
On multi-TB MySQL restores `xtrabackup --prepare` can run for hours between the basebackup download finishing and the database becoming available. From the outside the restore looks stuck — the bytes-downloaded gauge is pinned at 100% and there's no other signal.                 
                                                                                                                                                                                                                                                                                          
  This PR adds a `RestoreCoordinator.Phase.preparing_backup` entered during `restore_basebackup()` once `xbstream` extract is done and `xtrabackup --prepare` (followed by the short `--move-back`) starts running. The pct is derived from the InnoDB scan-line LSN advancing across     
  `[first observed scan LSN, last_lsn from xtrabackup_checkpoints]`. `last_lsn` is the redo stop LSN — using `to_lsn` would hit 100% before the redo tail is applied; the lower bound is the first observed scan LSN rather than `from_lsn`, which is `0` on full backups and would pin   
  the bar near the top of the range.                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                          
  `BasebackupRestoreOperation` parses `xtrabackup` output, derives the pct, and fires a callback when the integer pct changes (deduped at the pct level — at most ~101 emissions per run). The coordinator's callback updates `state["basebackup_prepare_progress"]` and sets             
  `Phase.preparing_backup`; the first call carries `pct=None` and is what flips the phase from `restoring_basebackup`.                                                                                                                                                                    
                                                                                                                                                                                                                                                                                          
  `/status/restore` exposes `basebackup_prepare_progress`, gated on the live phase so failure paths (`Phase.failed`, `Phase.failed_basebackup`) don't leak a stale pct. `Controller.is_safe_to_reload` now also blocks reload during `preparing_backup`; neither it nor                   
  `restoring_basebackup` is independently resumable, so a reload would discard hours of work. A `myhoard.basebackup_restore.xtrabackup_prepare_progress` gauge is emitted alongside the callback so dashboards can show in-flight progress without polling `/status/restore`.

### DIFF
--- a/myhoard/basebackup_restore_operation.py
+++ b/myhoard/basebackup_restore_operation.py
@@ -5,10 +5,11 @@ from .util import (
     find_extra_xtrabackup_executables,
     get_xtrabackup_version,
     parse_version,
+    parse_xtrabackup_info,
 )
 from contextlib import suppress
 from rohmu.util import increase_pipe_capacity, set_stream_nonblocking
-from typing import Dict, Final, Optional
+from typing import Callable, Dict, Final, Optional
 
 import base64
 import fnmatch
@@ -33,6 +34,11 @@ class BasebackupRestoreOperation:
 
     prepare_completed_re = re.compile(r"Shutdown completed; log sequence number (\d+)$")
 
+    # Canonical InnoDB scan line, emitted repeatedly during recovery; its LSN
+    # monotonically advances towards last_lsn and is the only signal we use for
+    # prepare progress.
+    prepare_lsn_re = re.compile(r"Doing recovery: scanned up to log sequence number (\d+)")
+
     def __init__(
         self,
         *,
@@ -46,6 +52,7 @@ class BasebackupRestoreOperation:
         target_dir: str,
         temp_dir: str,
         backup_tool_version: str | None = None,
+        prepare_progress_callback: Optional[Callable[..., None]] = None,
     ):
         self.current_file = None
         self.data_directory_size_end = None
@@ -57,7 +64,7 @@ class BasebackupRestoreOperation:
         self.mysql_config_file_name = mysql_config_file_name
         self.mysql_data_directory = mysql_data_directory
         self.number_of_files = 0
-        self.prepared_lsn = None
+        self.prepared_lsn: Optional[int] = None
         self.proc: subprocess.Popen[bytes] | None = None
         self.stats = stats
         self.stream_handler = stream_handler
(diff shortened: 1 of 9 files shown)


VitaliVinahradski (dismissed):
LGTM

VitaliVinahradski (dismissed):
LGTM

myhoard/basebackup_restore_operation.py
```diff
@@ -65,12 +73,55 @@ def __init__(
...
+        # truncate to the same pct. Reset on each prepare_backup call.
+        self._last_emitted_prepare_pct: Optional[int] = None
+        #: Fired when the xtrabackup --prepare subprocess is about to start (with
+        #: pct=None) and again whenever the derived pct changes during the run.
+        #: Called synchronously from the thread that invoked prepare_backup() —
+        #: _process_output_loop drains stdout/stderr in-line and calls
+        #: _process_prepare_output_line which emits this callback. Callers must
+        #: not block in the callback for long; the prepare output parser can't
+        #: make progress until it returns.
+        self.prepare_progress_callback: Optional[Callable[..., None]] = None
```
danielblasina-tech: would't be better to make it "private" and set the callback in the constructor?

myhoard/basebackup_restore_operation.py
```diff
@@ -65,12 +73,55 @@ def __init__(
...
+        # * prepare_scan_start_lsn is the first LSN observed in a
+        #   "Doing recovery: scanned up to log sequence number N" line. We don't use
+        #   from_lsn from the checkpoints here because it's 0 for every full backup
+        #   (it's the base-LSN for incrementals), which would pin pct at ~99% for
+        #   almost the entire prepare run — the actual redo replayed during --prepare
+        #   spans roughly the backup's duration, not the full history of the server.
+        # * prepare_current_lsn is the latest scan-line LSN seen.
+        self.prepare_last_lsn: Optional[int] = None
+        self.prepare_scan_start_lsn: Optional[int] = None
+        self.prepare_current_lsn: Optional[int] = None
```
danielblasina-tech: can be private?

myhoard/basebackup_restore_operation.py
```diff
@@ -33,6 +34,13 @@ class BasebackupRestoreOperation:
 
     prepare_completed_re = re.compile(r"Shutdown completed; log sequence number (\d+)$")
 
+    # Captures the currently-recovered LSN from the canonical InnoDB scan line, e.g.
+    #   "InnoDB: Doing recovery: scanned up to log sequence number 49787916061696"
+    # This is the only line we rely on for prepare progress: it is emitted
+    # repeatedly during recovery and its LSN monotonically advances towards last_lsn
+    # (the redo stop LSN). See prepare_last_lsn for why it's last_lsn and not to_lsn.
+    prepare_lsn_re = re.compile(r"Doing recovery:\s*scanned up to log sequence number\s*(\d+)")
```
danielblasina-tech: it is not important but why \s*? isn't always only one space

myhoard/basebackup_restore_operation.py
alexole: yes, but then most of the things can be private, I would just stick to the same pattern

myhoard/basebackup_restore_operation.py
alexole: 🤖

myhoard/basebackup_restore_operation.py
alexole: but my comment then contradicts the following new attributes, if you want I can make them provate

myhoard/basebackup_restore_operation.py
```diff
@@ -336,23 +389,103 @@ def _process_move_output_line(self, line: str, _stream_name: str) -> None:
...
+        match = self.prepare_lsn_re.search(line)
+        if not match:
+            return False
+        lsn = int(match.group(1))
+        if self._prepare_current_lsn is not None and lsn <= self._prepare_current_lsn:
+            return False
+        self._prepare_current_lsn = lsn
+        if self._prepare_scan_start_lsn is None:
+            self._prepare_scan_start_lsn = lsn
+        return True
```
danielblasina-tech: I think that this method should be in half, in specific this part is doing the same thing or a very similar thing that is doing in _process_prepare_completed_line, i would create a separate method for this part

myhoard/basebackup_restore_operation.py
```diff
@@ -336,23 +389,103 @@ def _process_move_output_line(self, line: str, _stream_name: str) -> None:
...
+            # int() truncation on a small range). Seed scan_start too if no scan
+            # line ever landed — collapsing the window to zero yields pct=100 from
+            # the property (handled by the span<=0 branch).
+            if self._prepare_last_lsn is not None and (
+                self._prepare_current_lsn is None or self._prepare_current_lsn < self._prepare_last_lsn
+            ):
+                self._prepare_current_lsn = self._prepare_last_lsn
+                if self._prepare_scan_start_lsn is None:
+                    self._prepare_scan_start_lsn = self._prepare_last_lsn
+                self._emit_prepare_progress(pct=self.prepare_progress_pct)
```
danielblasina-tech: I don't understand why we need to pass this property as parameter

danielblasina-tech (approved):
LGTM

PR #233, opened 2026-05-13
/backups/paused moved to /backup/settings

danielblasina-tech:
/backups/paused moved to /backup/settings

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -145,7 +145,7 @@ class Controller(threading.Thread):
         stream_to_be_purged: Optional[str]
         server_uuid: Optional[str]
         uploaded_binlogs: list
-        pause_backups_until: str | None
+        paused_backups_until: str | None
 
     def __init__(
         self,
@@ -236,7 +236,7 @@ class Controller(threading.Thread):
             "stream_to_be_purged": None,
             "server_uuid": None,
             "uploaded_binlogs": [],
-            "pause_backups_until": None,
+            "paused_backups_until": None,
         }
         self.state_dir = state_dir
         state_file = os.path.join(state_dir, "myhoard_controller_state.json")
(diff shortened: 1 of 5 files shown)


myhoard/controller.py
```diff
@@ -310,9 +310,9 @@ def mark_backup_preservation(self, stream_id: str, preserve_until: Optional[date
             current_requests[stream_id] = preserve_until.isoformat() if preserve_until else None
             self.state_manager.update_state(pending_preservation_requests=current_requests)
 
-    def pause_backups(self, until: datetime.datetime) -> None:
-        self.log.warning("Pausing backups until %s", until)
-        self.state_manager.update_state(pause_backups_until=until.isoformat())
+    def pause_backups(self, paused_until: datetime.datetime | None) -> None:
+        self.log.warning("Pausing backups until %s", paused_until)
```
alexole: It would be also great if we could log when backups are resumed as well.

danielblasina-tech: do you mean when paused_until is None?

alexole: exactly

danielblasina-tech: done

myhoard/web_server.py
```diff
@@ -89,21 +90,33 @@ async def backup_list(self, _request):
...
+            self.controller.pause_backups(paused_until=None)
+            return
 
-            return json_response({"success": True})
+        try:
+            paused_until_date = datetime.fromisoformat(paused_until)
+        except ValueError:
+            raise BadRequest("`paused_until` must be a valid iso-format datetime string.")
+
+        if paused_until_date < datetime.now(timezone.utc):
```
alexole, line 119: I think the comparison might end up in exception if the client sends no TZ because of comparison here with TZ aware date. And this will be 500 error then. Could you please check?

alexole (approved):

PR #234, opened 2026-05-18
Allow restoring into an empty existing data directory

SCedricThomas:
## What changed

RestoreCoordinator now allows restoring into an existing MySQL data directory when that directory is empty.

The previous behavior rejected any pre-existing path unconditionally. The new behavior keeps the safety check for existing content, but permits reuse of an already-created empty directory.

## Why

In some environments the target data directory may already exist before restore starts. Rejecting that case prevents restore even when the directory is safe to use.

This change keeps the guard against accidental overwrite by continuing to reject non-empty directories and non-directory paths.

## Tests

- added a unit test that allows restore when the target directory already exists and is empty
- added a unit test that still rejects restore when the target directory already exists and contains files

### DIFF
--- a/myhoard/restore_coordinator.py
+++ b/myhoard/restore_coordinator.py
@@ -501,7 +501,10 @@ class RestoreCoordinator(threading.Thread):
 
     def restore_basebackup(self) -> None:
         if os.path.exists(self.mysql_data_directory):
-            raise ValueError(f"MySQL data directory {self.mysql_data_directory!r} already exists")
+            if not os.path.isdir(self.mysql_data_directory):
+                raise ValueError(f"MySQL data directory {self.mysql_data_directory!r} already exists and is not a directory")
+            if os.listdir(self.mysql_data_directory):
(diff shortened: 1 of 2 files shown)


myhoard/restore_coordinator.py
```diff
@@ -501,7 +501,10 @@ def _run_basebackup_restore_operation(
 
     def restore_basebackup(self) -> None:
         if os.path.exists(self.mysql_data_directory):
-            raise ValueError(f"MySQL data directory {self.mysql_data_directory!r} already exists")
+            if not os.path.isdir(self.mysql_data_directory):
+                raise ValueError(f"MySQL data directory {self.mysql_data_directory!r} already exists and is not a directory")
+            if os.listdir(self.mysql_data_directory):
```
alexole: Thanks for your contribution.
Could you please replace costly `os.listdir`, which returns all the entities (it can be overkill if the dir has lots of them), with `os.scandir` and check only one first iteration?

alexole (approved):

PR #235, opened 2026-06-04
Recovering missing binlogs

danielblasina-tech:
Binlog scanner can now detect holes in binlogs
WebServer now has endpoint to update the index

### DIFF
--- a/myhoard/binlog_scanner.py
+++ b/myhoard/binlog_scanner.py
@@ -2,10 +2,12 @@
 from .append_only_state_manager import AppendOnlyStateManager
 from .state_manager import StateManager
 from .util import build_gtid_ranges, GtidRangeDict, read_gtids_from_log
+from pathlib import Path
 from typing import List, Optional, TypedDict
 
 import logging
 import os
+import re
 import threading
 import time
 
@@ -58,6 +60,7 @@ class BinlogScanner:
         self.state_manager = StateManager[BinlogScanner.State](state=self.state, state_file=state_file)
         self.stats = stats
         self.server_id = server_id
+        self.seen_missing_binlog = False
 
     @property
     def latest_complete_binlog_index(self) -> int:
@@ -70,7 +73,18 @@ class BinlogScanner:
         last_processed_at = time.time()
 
         next_index: int = self.state["next_index"]
+
         for _ in range(BINLOG_UPLOAD_BATCH_SIZE):
+            if not os.path.exists(self.build_full_name(next_index)):
+                if self.seen_missing_binlog:
+                    break
+                if self._has_missing_binlog(next_index):
+                    self.log.error("Binlog with index %s is missing, backup is broken", next_index)
+                    self.seen_missing_binlog = True
+                    break
+
+            self.seen_missing_binlog = False
+
             # We only scan completed files so expect the index following our next
             # index to be present as well
             if not os.path.exists(self.build_full_name(next_index + 1)):
@@ -119,6 +133,7 @@ class BinlogScanner:
                     total_binlog_count=self.state["total_binlog_count"] + len(added),
                     total_binlog_size=self.state["total_binlog_size"] + new_size,
                 )
+        self.stats.gauge_int("myhoard.binlog.missing", 1 if self.seen_missing_binlog else 0)
         # Send data points regardless of whether we got any new binlogs
         self.stats.gauge_int("myhoard.binlog.count", self.state["total_binlog_count"])
         self.stats.gauge_int("myhoard.binlog.size", self.state["total_binlog_size"])
@@ -171,3 +186,18 @@ class BinlogScanner:
         # Extensions are zero padded to be always at least 6 characters. I.e. file names
         # are prefix.000001, prefix.001000, prefix.100000, prefix.1000000, etc
         return f"{self.binlog_prefix}.{index:06}"
+
+    def _has_missing_binlog(self, binlog_index: int) -> int | None:
+        prefix_path = Path(self.binlog_prefix)
+        directory = prefix_path.parent
+        base_name = prefix_path.name
+        pattern = re.compile(rf"^{re.escape(base_name)}\.(\d+)$")
+
+        for path in directory.glob(f"{base_name}.*"):
+            match = pattern.match(path.name)
+            if match:
+                index = int(match.group(1))
+                if index > binlog_index:
+                    return True
+
+        return False
--- a/myhoard/web_server.py
+++ b/myhoard/web_server.py
@@ -6,6 +6,7 @@ from myhoard.backup_stream import BackupStream
 from myhoard.controller import Controller
 from myhoard.errors import BadRequest
 from myhoard.restore_coordinator import RestoreCoordinator
+from os import path
 from typing import Any
 
 import asyncio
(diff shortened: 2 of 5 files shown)


myhoard/web_server.py
```diff
@@ -264,6 +264,30 @@ async def status_update(self, request):
...
+    def _set_binlog_index(self, request):
+        try:
+            binlog_index = int(request.get("binlog_index"))
+        except (TypeError, ValueError):
+            raise BadRequest("'binlog_index' is not a valid integer")
+
+        if binlog_index < 1:
+            raise BadRequest("Min value for 'binlog_index' is 1")
+
+        self.controller.binlog_scanner.state["next_index"] = binlog_index
```
alexole: We should persist is using self.controller.binlog_scanner.state_manager.update_state, otherwise it will only be kept in memory

alexole: Also should probably be under `with self.controller.lock:`

myhoard/binlog_scanner.py
```diff
@@ -70,6 +70,11 @@ def scan_new(self, added_callback) -> List[BinlogInfo]:
         last_processed_at = time.time()
 
         next_index: int = self.state["next_index"]
+
+        if not os.path.exists(self.build_full_name(next_index)):
+            self.log.error("Binlog with index %s is missing, backup is broken", next_index)
+            return []
```
alexole: Shouldn't we reflect it properly in the metrics? Metrics are sent at the end of the method, so in this case will just be skipped.

alexole: BTW it would probably be good to emit a specific metric here, so the alert can be done based on that. And OPS can react.
Ideally we should also stop spamming with this log entry and write it only once? Otherwise will be too flooded on every scan.

myhoard/web_server.py
```diff
@@ -264,6 +264,32 @@ async def status_update(self, request):
...
+        except (TypeError, ValueError):
+            raise BadRequest("'binlog_index' is not a valid integer")
+
+        if binlog_index < 1:
+            raise BadRequest("Min value for 'binlog_index' is 1")
+
+        binlog_scanner = self.controller.binlog_scanner
+        with binlog_scanner.lock:
+            binlog_scanner.state_manager.update_state(next_index=binlog_index)
+        self.log.warning("Updating binlog index to %s", binlog_index)
```
alexole: nit: this message should be before the actual update, or should say "Updated"

myhoard/binlog_scanner.py
danielblasina-tech: Did prefer that logs the missing binlog constantly as is more noticeable, but ok

myhoard/binlog_scanner.py
```diff
@@ -70,7 +71,16 @@ def scan_new(self, added_callback) -> List[BinlogInfo]:
         last_processed_at = time.time()
 
         next_index: int = self.state["next_index"]
+
         for _ in range(BINLOG_UPLOAD_BATCH_SIZE):
+            if not os.path.exists(self.build_full_name(next_index)):
+                if not self.seen_missing_binlog:
+                    self.log.error("Binlog with index %s is missing, backup is broken", next_index)
+                    self.seen_missing_binlog = True
+                self.stats.gauge_int("myhoard.binlog.missing", 1)
```
alexole: it would be also good to send myhoard.binlog.missing = 0 when there is no gap anymore

myhoard/web_server.py
```diff
@@ -264,6 +264,32 @@ async def status_update(self, request):
...
+
+            return json_response({"success": True})
+
+    def _set_binlog_index(self, request):
+        try:
+            binlog_index = int(request.get("binlog_index"))
+        except (TypeError, ValueError):
+            raise BadRequest("'binlog_index' is not a valid integer")
+
+        if binlog_index < 1:
```
alexole, line 286: It would probably be nice to also check that file exists, it won't guarantee much but at least will be a sanity guard.
Also we need to be sure the new binlog_index is greater than the current value (latest_complete_binlog_index?), if it's not, we might actually break the state of the scanner, if I am not mistaken.

myhoard/web_server.py
```diff
@@ -264,6 +265,38 @@ async def status_update(self, request):
...
+            raise BadRequest("Min value for 'binlog_index' is 1")
```
danielblasina-tech, line 287: this is not needed anymore but doesn't hurt

myhoard/web_server.py
```diff
@@ -264,6 +265,38 @@ async def status_update(self, request):
...
+        binlog_scanner = self.controller.binlog_scanner
+
+        if binlog_index < binlog_scanner.latest_complete_binlog_index:
+            raise BadRequest("Binlog index cannot be decreased.")
```
alexole: All of those checks should be under the lock to make read and write safe:
```
with binlog_scanner.lock:
   if binlog_index < binlog_scanner.latest_complete_binlog_index:
...
```

myhoard/binlog_scanner.py
```diff
@@ -119,6 +129,7 @@ def scan_new(self, added_callback) -> List[BinlogInfo]:
                     total_binlog_count=self.state["total_binlog_count"] + len(added),
                     total_binlog_size=self.state["total_binlog_size"] + new_size,
                 )
+        self.stats.gauge_int("myhoard.binlog.missing", 0)
```
alexole: Is this correct? If I read the code correctly, it will emit 0 even if there is a gap right after emitting 1.

myhoard/binlog_scanner.py
```diff
@@ -171,3 +185,20 @@ def build_full_name(self, index):
...
+
+        prefix_path = Path(self.binlog_prefix)
+        directory = prefix_path.parent
+        base_name = prefix_path.name
+
+        found_indices = []
+        for path in directory.glob(f"{base_name}.*"):
+            index = int(path.suffix.lstrip("."))
+            if index >= binlog_index:
+                found_indices.append(index)
```
alexole: should be >, also if we find any index > binlog_index we can already return result, no need to keep on scanning and keep all of that in memory

alexole: If current file does not exist and there is any > current index, then we are certain there is a gap.

myhoard/binlog_scanner.py
```diff
@@ -171,3 +185,17 @@ def build_full_name(self, index):
...
+
+    def _has_missing_binlog(self, binlog_index: int) -> int | None:
+        if os.path.exists(self.build_full_name(binlog_index)):
+            return False
+
+        prefix_path = Path(self.binlog_prefix)
+        directory = prefix_path.parent
+        base_name = prefix_path.name
+
+        for path in directory.glob(f"{base_name}.*"):
```
alexole, line 196: Please replace with iglob, so it's an iterator, not a potentially massive list

myhoard/binlog_scanner.py
```diff
@@ -70,7 +72,18 @@ def scan_new(self, added_callback) -> List[BinlogInfo]:
...
         next_index: int = self.state["next_index"]
+
         for _ in range(BINLOG_UPLOAD_BATCH_SIZE):
+            if not self.seen_missing_binlog:
+                if self._has_missing_binlog(next_index):
+                    self.log.error("Binlog with index %s is missing, backup is broken", next_index)
+                    self.seen_missing_binlog = True
+                    break
+            elif not os.path.exists(self.build_full_name(next_index)):
+                break
```
alexole: You can remove os.path.exists(self.build_full_name(binlog_index)) check from _has_missing_binlog, because you are already calling it here, you can just reuse it and in this case it would look like
```python
            if not os.path.exists(self.build_full_name(next_index)):
                if self.seen_missing_binlog:
                    break
                if self._has_missing_binlog(next_index):
                    self.log.error("Binlog with index %s is missing, backup is broken", next_index)
                    self.seen_missing_binlog = True
                    break
```

myhoard/binlog_scanner.py
```diff
@@ -171,3 +185,17 @@ def build_full_name(self, index):
...
+    def _has_missing_binlog(self, binlog_index: int) -> int | None:
+        if os.path.exists(self.build_full_name(binlog_index)):
+            return False
+
+        prefix_path = Path(self.binlog_prefix)
+        directory = prefix_path.parent
+        base_name = prefix_path.name
+
+        for path in directory.glob(f"{base_name}.*"):
+            index = int(path.suffix.lstrip("."))
```
alexole: We should ignore any file which does not match our pattern, otherwise it will crash here with ValueError, or you could use tune pattern to something like  f"{base_name}.[0-9]*"

myhoard/binlog_scanner.py
danielblasina-tech: Path.glob should be an iterator

alexole: Ah, sorry, confused with glob.glob

alexole (approved):

PR #236, opened 2026-06-09
Fix myhoard for Python 3.14

orange-kao:
On Python 3.14, `self.loop = asyncio.get_event_loop() in myhoard/myhoard.py` will raise `RuntimeError: There is no current event loop in thread 'MainThread'`

This commit allows myhoard to run on Python 3.14 by using
```
self.loop = asyncio.new_event_loop()
asyncio.set_event_loop(self.loop)
````
instead

Make myhoard compatible with Python 3.14

Reference: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop
> Changed in version 3.14: Raises a [RuntimeError] if there is no current event loop.

### DIFF
--- a/myhoard/myhoard.py
+++ b/myhoard/myhoard.py
@@ -36,7 +36,7 @@ class MyHoard:
         self.controller = None
         self.is_running = True
         self.log = logging.getLogger(self.__class__.__name__)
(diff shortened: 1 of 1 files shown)


VitaliVinahradski (approved):
LGTM

PR #237, opened 2026-06-17
Fix invalid ISO 8601 in backup stream IDs

dogukancagatay:
`new_stream_id()` builds each identifier from a timezone-aware UTC
`datetime.isoformat()` already emits the `+00:00` offset, yet a literal
`Z` was appended on top. So every generated ID carried both an offset
and a trailing `Z`. That is not valid ISO 8601.

Drop the redundant `Z. New stream IDs are now well-formed:

  before: `2025-01-14T10:49:37.831286+00:00Z_<uuid>`
  after:  `2025-01-14T10:49:37.831286+00:00_<uuid>`

Existing IDs keep working. They are treated as opaque keys and
backups order by completed_at, not the embedded timestamp, so no
migration or backfill is needed.

### DIFF
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -553,7 +553,7 @@ class BackupStream(threading.Thread):
 
     @staticmethod
     def new_stream_id() -> str:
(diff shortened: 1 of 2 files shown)


VitaliVinahradski (approved):
LGTM

PR #238, opened 2026-06-26
Add estimate_memory xtrabackup setting for Smart memory estimation [MYC-106]

alexole:
## What

Adds an `estimate_memory` xtrabackup setting. When enabled, myhoard passes `--estimate-memory=ON` to `xtrabackup --backup`.

This is the **backup-side half** of Percona XtraBackup's [Smart memory estimation](https://docs.percona.com/percona-xtrabackup/8.0/smart-memory-estimation.html) (tech preview). The restore side — `--use-free-memory-pct` — is already supported (`BasebackupRestoreOperation`, driven by `restore_free_memory_percentage`), but xtrabackup **ignores** `--use-free-memory-pct` at prepare time unless the backup was taken with `--estimate-memory=ON`. So the two must be enabled together; this PR closes the gap.

## Why

The `xtrabackup --prepare` phase otherwise runs at the ~100MB default buffer, which is slow on large restores. Smart memory estimation lets prepare use a configurable share of free memory (Percona reports ~10–17x faster prepare in their environment). Enabling it requires turning on estimate-memory **at backup time**.

## Compatibility

- Default off → existing backups and callers unaffected.
- Feature is PXB tech preview; introduced in PXB 8.0.30-23, `--estimate-memory` option in 8.0.32-26. The restore side already version-gates `--use-free-memory-pct` to `>= 8.0.32`.

### DIFF
--- a/myhoard.json
+++ b/myhoard.json
@@ -75,6 +75,7 @@
         "copy_threads": 1,
         "compress_threads": 1,
         "encrypt_threads": 1,
+        "estimate_memory": false,
         "register_redo_log_consumer": false
     }
 }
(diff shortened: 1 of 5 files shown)


VitaliVinahradski (approved):
LGTM

PR #239, opened 2026-06-29
Bound per-cycle binlog purge to avoid LOCK_index write stalls

vladislav-manafov-aiven:
Reducing binlog_retention_period on a write-heavy service made the next 60s purge tick collect the entire newly-eligible backlog and issue one PURGE BINARY LOGS over all of it. MySQL runs that under LOCK_index, which every committing transaction also needs, so a large backlog stalls all writes for the full purge duration (hours / >1 TB).

Cap each purge cycle by file count and total bytes so a large backlog drains as many small purges across successive ticks instead of one long stall. The cap trims the tail of the already-computed contiguous oldest-first prefix, so all existing safety semantics (stream safety, min_binlog_age_before_purge, replication state, no-GTID handling) are preserved. At least one file is always purged to guarantee progress.

Also emit myhoard.binlog.purged_per_cycle.{files,bytes} so the cap can be verified.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -618,10 +618,17 @@ class Controller(threading.Thread):
 
         Returns the longest prefix that satisfies both max_files_per_cycle and max_bytes_per_cycle. At least one
         binlog is always kept (even if it alone exceeds max_bytes_per_cycle) so that purging always makes forward
-        progress. A cap of None (or absent) means unbounded for that dimension.
+        progress. A cap that is None, absent or non-positive means unbounded for that dimension.
         """
         max_files = purge_settings.get("max_files_per_cycle")
         max_bytes = purge_settings.get("max_bytes_per_cycle")
+        # A cap only takes effect when it is a positive number. Normalize any non-positive value to None (no limit)
+        # so both dimensions behave consistently: otherwise max_bytes_per_cycle=0 would still purge a single file
+        # due to the forward-progress guarantee below, while max_files_per_cycle=0 would purge nothing at all.
+        if max_files is not None and max_files <= 0:
+            max_files = None
+        if max_bytes is not None and max_bytes <= 0:
+            max_bytes = None
         if (max_files is None and max_bytes is None) or not binlogs_to_purge:
             return binlogs_to_purge
 
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -605,7 +605,47 @@ class Controller(threading.Thread):
                     binlogs_to_purge.append(binlog)
                 else:
                     break
+        # Bound the per-cycle purge so a large eligible backlog is drained as many small purges across successive
+        # ticks instead of a single PURGE BINARY LOGS covering the whole backlog (which holds LOCK_index, stalling
+        # all writes, for the entire duration). binlogs_to_purge is a contiguous oldest-first prefix and any
+        # shorter prefix is equally safe to purge, so we simply truncate it to satisfy whichever cap is hit first.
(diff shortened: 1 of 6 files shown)


myhoard/controller.py
```diff
@@ -605,8 +605,48 @@ def collect_binlogs_to_purge(
...
+    @staticmethod
+    def _cap_binlogs_to_purge(binlogs_to_purge, *, purge_settings, log):
+        """Trim a contiguous oldest-first list of purgeable binlogs to the per-cycle caps.
+
+        Returns the longest prefix that satisfies both max_files_per_cycle and max_bytes_per_cycle. At least one
+        binlog is always kept (even if it alone exceeds max_bytes_per_cycle) so that purging always makes forward
+        progress. A cap of None (or absent) means unbounded for that dimension.
+        """
+        max_files = purge_settings.get("max_files_per_cycle")
+        max_bytes = purge_settings.get("max_bytes_per_cycle")
```
alexole, line 624: Can we also update the README with the new settings and have some reasonable defaults in myhoard.json?

vladislav-manafov-aiven: Updated

myhoard/controller.py
```diff
@@ -605,8 +605,48 @@ def collect_binlogs_to_purge(
...
+        if (max_files is None and max_bytes is None) or not binlogs_to_purge:
+            return binlogs_to_purge
+
+        capped = []
+        total_bytes = 0
+        for binlog in binlogs_to_purge:
+            if max_files is not None and len(capped) >= max_files:
```
alexole, line 638: I think we should treat 0 values as default - no limit, otherwise it would work unexpectedly. Also now seems that behaviour is not consistent between both settings when they are 0. E.g. when max_files_per_cycle is 0, it will halt without even purging anything. However it seems it will purge when max_bytes_per_cycle=0. Or we should set values to the least minimal values we allow - 1 in case 0 is set.

vladislav-manafov-aiven: That is right, an inconsistency. Fixed that. Zero value is treated as None

myhoard/controller.py
```diff
@@ -605,8 +605,55 @@ def collect_binlogs_to_purge(
...
+        Returns the longest prefix that satisfies both max_files_per_cycle and max_bytes_per_cycle. At least one
+        binlog is always kept (even if it alone exceeds max_bytes_per_cycle) so that purging always makes forward
+        progress. A cap that is None, absent or non-positive means unbounded for that dimension.
+        """
+        max_files = purge_settings.get("max_files_per_cycle")
+        max_bytes = purge_settings.get("max_bytes_per_cycle")
+        # A cap only takes effect when it is a positive number. Normalize any non-positive value to None (no limit)
+        # so both dimensions behave consistently: otherwise max_bytes_per_cycle=0 would still purge a single file
+        # due to the forward-progress guarantee below, while max_files_per_cycle=0 would purge nothing at all.
+        if max_files is not None and max_files <= 0:
```
alexole, line 628: Now we are implicitly turning negative values into None, which I don't think is ideal.
IMO we should throw validation error in case this happens. And when I think about it now, if we add validation, it would be more reasonable to actually reject everything <=0 and leave None as default behaviour. WDYT?

alexole: Or at least then a log message that using None value - no limit.

alexole: Alright, it will be noisy then. Ignore it.

alexole (approved):

alexole:
Please create a follow up to reduce interval when throttled to 1 second or even less, otherwise the rate of binlogs generation might be higher then removing and there is a risk of filling up the disk.

PR #240, opened 2026-07-07
Drain capped binlog purge backlog promptly with optional inter-chunk delay

vladislav-manafov-aiven:
The per-cycle caps keep each PURGE small, but purging only runs once per
purge_interval (60s), so a big backlog cleared just one chunk a minute — 
hours or days for a large one.

Now, when a capped chunk leaves more to purge, we let the next chunk run on
the following control-loop tick (~1s) instead of waiting a full interval. The
backlog drains steadily while each PURGE still holds LOCK_index only briefly,
and the ~1s gap gives ordinary transactions room to commit in between. We only
speed up after a chunk actually purged, so a failed PURGE just retries at the
normal pace.

### DIFF
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -609,16 +609,24 @@ class Controller(threading.Thread):
         # ticks instead of a single PURGE BINARY LOGS covering the whole backlog (which holds LOCK_index, stalling
         # all writes, for the entire duration). binlogs_to_purge is a contiguous oldest-first prefix and any
         # shorter prefix is equally safe to purge, so we simply truncate it to satisfy whichever cap is hit first.
-        binlogs_to_purge = cls._cap_binlogs_to_purge(binlogs_to_purge, purge_settings=purge_settings, log=log)
-        return binlogs_to_purge, bool(only_binlogs_without_gtids or only_binlogs_that_are_too_new)
+        binlogs_to_purge, more_purgeable_remaining = cls._cap_binlogs_to_purge(
+            binlogs_to_purge, purge_settings=purge_settings, log=log
+        )
+        return (
+            binlogs_to_purge,
+            bool(only_binlogs_without_gtids or only_binlogs_that_are_too_new),
+            more_purgeable_remaining,
+        )
 
     @staticmethod
     def _cap_binlogs_to_purge(binlogs_to_purge, *, purge_settings, log):
         """Trim a contiguous oldest-first list of purgeable binlogs to the per-cycle caps.
 
-        Returns the longest prefix that satisfies both max_files_per_cycle and max_bytes_per_cycle. At least one
-        binlog is always kept (even if it alone exceeds max_bytes_per_cycle) so that purging always makes forward
-        progress. A cap that is None, absent or non-positive means unbounded for that dimension.
+        Returns a ``(capped, was_capped)`` tuple where ``capped`` is the longest prefix that satisfies both
+        max_files_per_cycle and max_bytes_per_cycle and ``was_capped`` indicates whether anything was trimmed
+        (i.e. a safe backlog still remains after this cycle). At least one binlog is always kept (even if it alone
+        exceeds max_bytes_per_cycle) so that purging always makes forward progress. A cap that is None, absent or
+        non-positive means unbounded for that dimension.
         """
         max_files = purge_settings.get("max_files_per_cycle")
         max_bytes = purge_settings.get("max_bytes_per_cycle")
@@ -630,7 +638,7 @@ class Controller(threading.Thread):
         if max_bytes is not None and max_bytes <= 0:
             max_bytes = None
         if (max_files is None and max_bytes is None) or not binlogs_to_purge:
-            return binlogs_to_purge
+            return binlogs_to_purge, False
 
         capped = []
         total_bytes = 0
@@ -643,7 +651,8 @@ class Controller(threading.Thread):
             capped.append(binlog)
             total_bytes += binlog_size
 
-        if len(capped) < len(binlogs_to_purge):
+        was_capped = len(capped) < len(binlogs_to_purge)
+        if was_capped:
             log.info(
                 "Capping per-cycle purge to %s/%s binlogs (%s bytes); max_files_per_cycle=%s max_bytes_per_cycle=%s",
                 len(capped),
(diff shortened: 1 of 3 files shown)


myhoard/controller.py
```diff
@@ -1204,8 +1217,15 @@ def _get_iteration_sleep(self):
         # possible to reduce downtime
         if self.mode == self.Mode.promote:
             return self.iteration_sleep / 10.0
-        else:
-            return self.iteration_sleep
+        # While draining a large capped binlog purge backlog, pace the next chunk by the configured per-cycle delay
+        # (non-blocking: the loop just waits this long before the next iteration, so other controller work still runs
+        # between chunks). A delay of zero/absent falls through to the default iteration sleep, but the finally block
+        # in _purge_old_binlogs additionally wakes the loop immediately in that case for sub-second draining.
+        if self._draining_capped_backlog:
```
alexole: This lengthens the shared run-loop sleep, so during a drain a large `delay_between_purge_chunks_ms` also paces binlog rotation and local binlog processing (same loop), not just purge chunks. That contradicts the README's "does not hold up other myhoard work (binlog rotation...)". Fine for sub-second values (< `iteration_sleep`), but nothing bounds it. Either tighten the doc to say it paces the whole controller loop and is meant for small values, or drive the delay off `binlogs_purged_at` so `iteration_sleep` stays constant.

myhoard/controller.py
```diff
@@ -1655,6 +1675,10 @@ def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
         # This is reported in another metric to indicate time since we could've purged binlogs but didn't because
         # there was not sufficient info to make a decision (due to system being idle and no changes happening)
         last_could_have_purged = self.state["last_could_have_purged"]
+        # When a large eligible backlog is drained in per-cycle capped chunks we must not wait a full purge_interval
+        # between chunks or draining terabytes would take many hours. This is set when a chunk was purged but more
+        # safely-purgeable binlogs remain, so the interval gate is skipped and the next chunk runs on the next tick.
+        more_purgeable_remaining = False
```
alexole, line 1688: `_draining_capped_backlog` is only written in the finally, which the top-of-function guard (`if not enabled or interval gate closed: return`) skips. If purge is disabled or `purge_interval` is raised mid-drain, the flag stays `True` and a large delay keeps pacing the whole loop indefinitely. Simplest fix is to reset it before that early return:

```python
if not purge_settings["enabled"] or time.time() - self.state["binlogs_purged_at"] < purge_settings["purge_interval"]:
    self._draining_capped_backlog = False
    return
```

myhoard/controller.py
```diff
@@ -1723,12 +1747,27 @@ def _purge_old_binlogs(self, *, mysql_maybe_not_running=False):
...
-                binlogs_purged_at=current_time,
+                binlogs_purged_at=binlogs_purged_at,
                 last_binlog_purge=last_purge,
                 last_could_have_purged=last_could_have_purged,
             )
 
+            # Pace the drain via the run loop (non-blocking) rather than sleeping in this thread, so binlog rotation,
+            # new-binlog handoff and other controller work still run between chunks. With no configured delay we wake
+            # the loop immediately for sub-second draining; with a delay _get_iteration_sleep waits that long instead.
+            self._draining_capped_backlog = more_purgeable_remaining
```
alexole: `more_purgeable_remaining` is computed at collection time and the finally acts on it even when the PURGE early-returned on error (`ERR_TIMEOUT`/`ERR_CANNOT_CONNECT`/`ERR_BACKUP_IN_PROGRESS`). So on failure we rewind `binlogs_purged_at` and (default delay) set `wakeup_event`, turning what used to be a once-per-interval retry into a tight retry loop while e.g. a backup holds the lock. Suggest only fast-retrying when a chunk actually succeeded. (The comment at ~1678, "set when a chunk was purged," reads as if that's already the case.)

test/test_controller.py
```diff
@@ -1634,7 +1643,37 @@ def test_cap_binlogs_to_purge_non_positive_caps_mean_unbounded():
...
+        mode=mode,
+        iteration_sleep=iteration_sleep,
+        _draining_capped_backlog=draining,
+        binlog_purge_settings={"delay_between_purge_chunks_ms": delay_ms},
+        Mode=Controller.Mode,
+    )
+    return Controller._get_iteration_sleep(stub)  # pylint: disable=protected-access
+
+
+def test_get_iteration_sleep_paces_capped_drain_by_configured_delay():
```
alexole: These cover the tuple plumbing and `_get_iteration_sleep`, but not the actual new behavior in `_purge_old_binlogs`'s finally: rewinding `binlogs_purged_at` and setting `wakeup_event` when `more_purgeable_remaining`, and leaving them alone when not. Worth a focused test (mock the purge) to lock in that contract, including the failure-path behavior noted above.

alexole (commented):
A few notes below. Approach is right and the code is clean; the two I'd want addressed are the doc-vs-implementation mismatch for large delays and a focused test for the new finally-block behavior. The failure-path retry and the stale flag are worth a guard.

test/conftest.py
```diff
@@ -398,6 +398,7 @@ def fixture_myhoard_config(default_backup_site, mysql_master, session_tmpdir):
             "purge_interval": 60,
             "max_files_per_cycle": 50,
             "max_bytes_per_cycle": 5368709120,
+            "delay_between_purge_chunks_ms": 0,
```
alexole: Two things I'd change on delay_between_purge_chunks_ms.

1) The ms unit is misleading because the run loop only wakes on ~1s ticks, so any positive delay rounds up to a whole second. 1500ms actually waits ~2s (the +1s tick is too early, so it fires at the +2s tick), and everything in 1..1000ms collapses to ~1s. The millisecond precision doesn't exist. I'd rename it to seconds, integer, so the unit matches what the loop can actually deliver.

2) Also the default shouldn't be 0. Back-to-back chunks re-create the sustained LOCK_index pressure the caps are meant to relieve, just in smaller bites, so 0 doesn't leave room for committing transactions between purges. Default of 1 second gives real breathing room and still drains ~50 files/s at the default max_files_per_cycle=50, way faster than any realistic binlog generation rate, so no accumulation risk.

If sub-second pacing is ever genuinely wanted that's a separate change: wake the loop after the delay instead of letting it free-run on the tick. But for lock relief between chunks, second granularity is the honest and sufficient model.

alexole: As for the 2 you can keep 0 if you want, this is a bit philosophical, if we want to be close to the previous behaviour, still this is an improvement as there will be some gaps between the purges.

vladislav-manafov-aiven: I removed the `delay_between_purge_chunks_ms` option. Relying onto main cycle tick now (1s)

alexole:
Could you change the description of PR to reflect the changes? Otherwise LGTM

alexole (approved):

PR #242, opened 2026-07-23
Mark backup as broken

danielblasina-tech:
Sometimes backups can be broken, this feature adds a new endpoint for backups to mark or unmark backup as broken

### DIFF
--- a/README.md
+++ b/README.md
@@ -648,6 +648,41 @@ This endpoint manages binlogs
 }
 ```
 
+## PATCH /backup/{stream_id}
+
+This endpoint updates settings of a single, already completed backup.
+
+```
+{
+  "broken": "{bool}",
+  "expire_after_seconds": "{number}",
+}
+```
+
+**stream_id**
+
+Identifier of the backup to update.
+
+### Allowed parameters are:
+
+**broken** (bool) - Marks the backup as broken (`true`) or clears a previous broken
+mark (`false`). A broken backup is skipped when picking a backup to restore from. The
+request is queued and applied asynchronously by the controller.
+
+**expire_after_seconds** (number, optional) - Number of seconds to wait for the
+`broken` change to be applied before responding. Without this parameter the endpoint
+returns as soon as the request has been queued, without waiting for it to be applied.
+If the change has not been applied within the given time, or applying it failed, the
+endpoint responds with an error.
+
+### Response
+
+```
+{
+  "success": true
+}
+```
+
 ## PUT /backup/{stream_id}/preserve
 
 Create a request for updating the preservation status of a backup. Request body must be like this:
--- a/myhoard/backup_stream.py
+++ b/myhoard/backup_stream.py
@@ -32,6 +32,7 @@ from rohmu import errors as rohmu_errors
 from rohmu.compressor import CompressionStream
 from rohmu.encryptor import EncryptorStream
 from rohmu.errors import FileNotFoundFromStorageError
+from rohmu.object_storage.base import BaseTransfer
 from rohmu.object_storage.s3 import S3Transfer
 from socket import gaierror
 from socks import GeneralProxyError, ProxyConnectionError
@@ -51,7 +52,6 @@ import uuid
 
 if TYPE_CHECKING:
     from .statsd import StatsClient
-    from rohmu.object_storage.base import BaseTransfer
 
 
 BINLOG_BUCKET_SIZE = 500
@@ -428,33 +428,45 @@ class BackupStream(threading.Thread):
         for binlog in binlogs:
             yield binlog
 
-    def mark_as_broken(self) -> None:
-        if self.state["broken_info"]:
-            self.log.warning("Stream %s marked as broken multiple times", self.stream_id)
-            return
-
+    def _mark_as_broken(self, key: str, file_storage: BaseTransfer):
         self.stop()
-        self.log.info("Marking stream %s as broken.", self.stream_id)
 
         broken_info = {
             "broken_at": time.time(),
             "server_id": self.server_id,
         }
-        # Write the broken info to local state too
         self.state_manager.update_state(broken_info=broken_info)
 
-        try:
-            key = self._build_full_name("broken.json")
-            data = json.dumps(broken_info)
-            metadata = make_fs_metadata(broken_info)
+        data = json.dumps(broken_info)
+        metadata = make_fs_metadata(broken_info)
+        file_storage.store_file_from_memory(key, data.encode("utf-8"), metadata=metadata)
 
+    def _mark_as_non_broken(self, key: str, file_storage: BaseTransfer):
+        self.state_manager.update_state(broken_info=None)
+        file_storage.delete_key(key=key)
+
+    def mark_as_broken(self, broken: bool) -> Tuple[bool, Optional[str]]:
+        try:
+            self.log.info("Marking stream=%s to broken=%s", self.stream_id, broken)
             file_storage = self.file_storage_setup_fn()
-            file_storage.store_file_from_memory(key, data.encode("utf-8"), metadata=metadata)
+            key = self._build_full_name("broken.json")
+            if any(file_storage.iter_key(key=key, include_key=True)) == broken:
+                error = f"Error stream={self.stream_id} was marked as broken={broken} multiple times"
+                self.log.warning(error)
+                return False, error
+
+            if broken:
+                self._mark_as_broken(key=key, file_storage=file_storage)
+            else:
+                self._mark_as_non_broken(key=key, file_storage=file_storage)
+            return True, None
         except Exception as ex:  # pylint: disable=broad-except
-            self.log.exception("Failed to create broken.json")
+            error = f"Failed to set stream={self.stream_id} to broken={broken}"
+            self.log.exception(error)
             self.stats.unexpected_exception(ex=ex, where="BackupStream.mark_as_broken")
             self.state_manager.update_state(remote_write_errors=self.state["remote_write_errors"] + 1)
             self.stats.increase("myhoard.remote_write_errors")
+            return False, error
 
     def mark_as_closed(self) -> None:
         if self.state["closed_info"]:
--- a/myhoard/controller.py
+++ b/myhoard/controller.py
@@ -45,6 +45,7 @@ import pymysql
 import re
 import threading
 import time
+import uuid
 
 ERR_CANNOT_CONNECT = 2003
 ERR_BACKUP_IN_PROGRESS = 4085
@@ -100,6 +101,19 @@ def sort_completed_backups(backups: List[Backup], reverse: bool = False) -> List
     return sorted((backup for backup in backups if backup["completed_at"]), key=key, reverse=reverse)
 
 
+class BrokenBackupResult(TypedDict):
+    success: bool
+    error: str | None
+    expires_at: str
+
+
+class BrokenBackupRequest(TypedDict):
+    request_id: str
+    stream_id: str
+    broken: bool
+    expires_at: str | None
+
+
 class Controller(threading.Thread):
     """Main logic controller for the service. This drives the individual handlers like
     BackupStream, BinlogScanner and RestoreCoordinator as well as provides state info
@@ -146,6 +160,8 @@ class Controller(threading.Thread):
         server_uuid: Optional[str]
         uploaded_binlogs: list
         paused_backups_until: str | None
+        broken_backup_requests: list[BrokenBackupRequest]
+        broken_backup_results: dict[str, BrokenBackupResult]
 
     def __init__(
         self,
@@ -237,6 +253,8 @@ class Controller(threading.Thread):
             "server_uuid": None,
             "uploaded_binlogs": [],
             "paused_backups_until": None,
+            "broken_backup_requests": [],
+            "broken_backup_results": {},
         }
         self.state_dir = state_dir
         state_file = os.path.join(state_dir, "myhoard_controller_state.json")
@@ -1241,7 +1259,7 @@ class Controller(threading.Thread):
             raise KeyError(f"Stream {stream_id} not found in backups")
         return backup["site"]
 
-    def get_backup_by_stream_id(self, stream_id: str):
+    def get_backup_by_stream_id(self, stream_id: str) -> Backup | None:
         for backup in self.state["backups"]:
             if backup["stream_id"] == stream_id:
                 return backup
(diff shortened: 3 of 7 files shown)


myhoard/web_server.py
```diff
@@ -303,6 +287,29 @@ def _set_binlog_index(self, request):
...
 
+    async def _update_resource(self, name: str, request: web.Request, actions: dict[str, Callable[[dict[str, Any]], None]]):
+        resource_params = actions.keys()
+        with self._handle_request(name=name):
+            body = await self._get_request_json(request)
+            if len(body.keys() & resource_params) == 0:
+                raise BadRequest(f"No valid resource parameters were specified, valid params are: {resource_params}")
+
+            for key in body.keys():
+                actions[key](body)
```
VitaliVinahradski: Behavior change: unknown keys used to be ignored, now `{"paused_until": ..., "typo": 1}` passes the guard and `actions["typo"]` raises KeyError, which turns into a 500. And since keys apply in body order, the valid key may already be applied when it fails. I would iterate `body.keys() & resource_params` to keep the old semantics, or 400 on unknown keys before applying anything.

myhoard/backup_stream.py
```diff
@@ -428,28 +428,35 @@ def iterate_remote_binlogs(self, *, reverse: bool = False) -> Iterator[BinlogInf
...
             yield binlog
 
-    def mark_as_broken(self) -> None:
-        if self.state["broken_info"]:
+    def mark_as_broken(self, broken: bool) -> None:
+        if bool(self.state["broken_info"]) == broken:
             self.log.warning("Stream %s marked as broken multiple times", self.stream_id)
             return
 
         self.stop()
```
VitaliVinahradski, line 432: `stop()` now runs on the unmark path too. It happens to be harmless in the current implementation but I would only stop when `broken=True`.

myhoard/backup_stream.py
```diff
@@ -428,28 +428,35 @@ def iterate_remote_binlogs(self, *, reverse: bool = False) -> Iterator[BinlogInf
         for binlog in binlogs:
             yield binlog
 
-    def mark_as_broken(self) -> None:
-        if self.state["broken_info"]:
+    def mark_as_broken(self, broken: bool) -> None:
+        if bool(self.state["broken_info"]) == broken:
             self.log.warning("Stream %s marked as broken multiple times", self.stream_id)
```
VitaliVinahradski: The log lines in this method still assume the broken=True direction. Unmarking a stream that is not broken logs "marked as broken multiple times", a successful unmark logs "Marking stream ... as broken.", and a failed `delete_key` logs "Failed to create broken.json".

myhoard/controller.py
```diff
@@ -1345,16 +1345,19 @@ def _handle_mode_restore(self):
...
+        broken_backup = self.get_backup_by_stream_id(stream_id=stream_id)
 
         if not broken_backup:
             raise Exception(
-                f'Stream {failed_stream_id} to be marked as broken not found in completed backups: {self.state["backups"]}'
+                f'Stream {stream_id} to be marked as broken not found in completed backups: {self.state["backups"]}'
             )
 
-        self._build_backup_stream(broken_backup).mark_as_broken()
+        self._build_backup_stream(broken_backup).mark_as_broken(broken=broken)
```
VitaliVinahradski, line 1438: One thing to double check: this builds a second BackupStream for a stream_id that may also be live in `self.backup_streams` (the newest completed backup is typically still streaming binlogs). Might be worth routing through the live stream object when one exists.

VitaliVinahradski (changes requested):

myhoard/backup_stream.py
danielblasina-tech: now it logs correctlt

myhoard/controller.py
danielblasina-tech: can you explain better?

VitaliVinahradski: Sure. The check in mark_as_broken (`bool(self.state["broken_info"]) == broken`) reads the node-local state file, but the flag that actually matters is the remote broken.json: get_backup_list builds `backup["broken_at"]` from storage metadata and that is what restore selection uses. Those two can disagree easily:

1. This endpoint builds a second BackupStream for the stream_id, but for the newest completed backup the controller usually still has a live BackupStream streaming binlogs, pointing at the same state_file. StateManager keeps state in memory and rewrites the whole file on every update_state. So PATCH broken=true writes broken_info to the file and uploads broken.json, and moments later the live stream does a routine update_state (pending_binlogs, last_remote_state_check, anything) and rewrites the file from its own in-memory copy where broken_info is still empty. Local state now says not broken, remote says broken.

2. Same disagreement if the backup was marked broken on a different node. The failed-restore path marks it on whatever node was restoring, so this node's local state file never saw the mark, or does not exist at all, in which case the fresh StateManager just creates it with an empty broken_info.

Once local and remote disagree, PATCH {"broken": false} is a permanent no-op: the fresh stream loads local state, sees broken_info empty, hits the early-return guard, returns success, and never deletes the remote broken.json. The backup stays broken for restore selection and you cannot unmark it through the API.

So I would base the guard on remote state (e.g. `backup["broken_at"]` from the controller's backup list) instead of the fresh object's local state file, and reuse the live stream object when one exists so local state stays consistent.

VitaliVinahradski:
One more thing I noticed on the new endpoint: PATCH /backup/{stream_id} does the object storage I/O directly in the request handler. `_set_backup_broken` calls `controller.mark_stream_as_broken`, which runs `list_path` plus `store_file_from_memory`/`delete_key` inline, and the handler runs on the aiohttp event loop, so every other HTTP request is stalled while we wait on object storage.

The pattern elsewhere is to defer this to the controller thread: `backup_preserve` only records a pending request in local state and `_handle_pending_preservation_requests` does the actual storage work from the controller's own thread. There is also a comment in `mark_as_closed` about writing local state only, "to ensure file storage operation happens from appropriate thread". I would do the same here, record the request and let the controller process it.

https://github.com/Aiven-Open/myhoard/blob/4a18e5063a0c8f388d27fad1ebb9f892851615a8/myhoard/web_server.py#L306-L312

myhoard/controller.py
VitaliVinahradski, line 1285: PATCH with marking stream as broken in idle/observe/restore/promote never get processed

myhoard/backup_stream.py
```diff
@@ -428,30 +428,39 @@ def iterate_remote_binlogs(self, *, reverse: bool = False) -> Iterator[BinlogInf
...
+
+            if broken:
+                self._mark_as_broken(key=key, file_storage=file_storage)
+            else:
+                self._mark_as_non_broken(key=key, file_storage=file_storage)
 
-            file_storage = self.file_storage_setup_fn()
-            file_storage.store_file_from_memory(key, data.encode("utf-8"), metadata=metadata)
         except Exception as ex:  # pylint: disable=broad-except
-            self.log.exception("Failed to create broken.json")
```
VitaliVinahradski, line 454: it looks like failure will return `success: True` when object storage fail to crate broken.json

danielblasina-tech: yes it is expected, the response contains that was processed, doesn't tell if was successful

myhoard/controller.py
danielblasina-tech: only master can do operations on the backup, i can prevent adding to the `broken_backup_requests` if you think that is needed

myhoard/controller.py
```diff
@@ -1345,16 +1348,43 @@ def _handle_mode_restore(self):
...
-                f'Stream {failed_stream_id} to be marked as broken not found in completed backups: {self.state["backups"]}'
+                f'Stream {stream_id} to be marked as broken not found in completed backups: {self.state["backups"]}'
             )
+        return broken_backup
 
-        self._build_backup_stream(broken_backup).mark_as_broken()
+    def request_mark_stream_as_broken(self, stream_id: str, broken: bool) -> None:
+        with self.lock:
+            self._check_existing_streams(stream_id=stream_id)
+            requests = self.state["broken_backup_requests"]
```
VitaliVinahradski: I think we need to copy with 
```python
  requests = dict(self.state["broken_backup_requests"])
```
to avoid mutating live object which is prone to errors

see how pending_preservation_requests are handled in myhoard/controller.py:318 and myhoard/controller.py:1851

myhoard/controller.py
VitaliVinahradski: yes, but other modes should reject PATCH and not hang indefinitely

myhoard/controller.py
```diff
@@ -1345,16 +1348,43 @@ def _handle_mode_restore(self):
                 restore_options={},
             )
 
-    def _mark_failed_restore_backup_as_broken(self) -> None:
-        failed_stream_id = self.state["restore_options"]["stream_id"]
-        broken_backup = self.get_backup_by_stream_id(stream_id=failed_stream_id)
+    def _handle_mark_streams_as_broken(self):
+        with self.lock:
```
VitaliVinahradski, line 1368: Too much work inside lock blocking main event loop. `_handle_pending_preservation_requests` already has the shape I would copy here: take the lock to pick the next item, do the remote work outside it, then take it again to write.

prior art:
https://github.com/Aiven-Open/myhoard/blob/168d93e817dbf9d363b6b219144d8defd694f67b/myhoard/controller.py#L1839-L1856

test/test_controller.py
```diff
@@ -2374,6 +2374,22 @@ def test_count_non_broken_backups(
     assert Controller.count_non_broken_backups(controller.state["backups"]) == expected_result
 
 
+def test_request_mark_stream_as_broken_and_handle_mark_streams_as_broken(master_controller):
+    controller, _ = master_controller
+    add_fake_backup(controller, "1234")
+
+    controller.request_mark_stream_as_broken(stream_id="1234", broken=True)
+    assert controller.state["broken_backup_requests"] == {"1234": True}
```
VitaliVinahradski: Neither this test nor the web server one can catch a persistence problem here, since every assertion reads in-memory controller.state.  From reading the code I'd expect a disk-level assertion to fail. Worth probing with a reload after the enqueue:

```python
 controller.state_manager.read_state()
 assert controller.state["broken_backup_requests"] == {"1234": True}
```

add permanent test in its own test case.

myhoard/backup_stream.py
danielblasina-tech: now it returns the error to client

myhoard/controller.py
danielblasina-tech: solved

test/test_controller.py
danielblasina-tech: did create test test_request_mark_stream_as_broken_persists_to_disk for this

myhoard/controller.py
```diff
@@ -1345,16 +1364,75 @@ def _handle_mode_restore(self):
...
+            with self.lock:
+                broken_backup_request = self.state["broken_backup_requests"][0]
+            stream_id, request_id, expires_at, broken = broken_backup_request
+            broken_backup = self.get_backup_by_stream_id(stream_id=stream_id)
+            if not broken_backup:
+                error = f"Stream {stream_id} does not exist anymore"
+                logging.info("Skipping marking stream=%s to broken=%s as stream does not exist anymore", stream_id, broken)
+                succes = False
+            else:
+                succes, error = self._build_backup_stream(broken_backup).mark_as_broken(broken=broken)
```
VitaliVinahradski: The fix is small: look up `self.backup_streams` for the stream_id and reuse that instance when present, falling back to `_build_backup_stream` only for streams this node isn't running.

Failure scenario: operator marks the newest backup broken while its stream is live, the throwaway reverts the stream's file, the process restarts before the live stream flushes again, the stream comes back missing binlogs that the scanner has already moved past, and those binlogs are never uploaded. The result is a gap in the binlog chain, so PITR to any target after the gap silently fails for that stream.

Sketch, untested:

```python
live_stream = next((s for s in self.backup_streams if s.stream_id == stream_id), None)
succes, error = (live_stream or self._build_backup_stream(broken_backup)).mark_as_broken(broken=broken)
```

myhoard/web_server.py
```diff
@@ -303,6 +287,58 @@ def _set_binlog_index(self, request):
...
+        broken = body.get("broken")
+        if not isinstance(broken, bool):
+            raise BadRequest("'broken' is not a valid bool")
+        expire_after_seconds = body.get("expire_after_seconds")
+        request_id = self.controller.request_mark_stream_as_broken(
+            stream_id=stream_id, broken=broken, expire_after_seconds=expire_after_seconds
+        )
+
+        if expire_after_seconds:
+            if not isinstance(expire_after_seconds, (int, float)):
```
VitaliVinahradski: nit: validation happens after use.  Moving the isinstance check above the call fixes it. User will see 500 error on wrong values

README.md
```diff
@@ -648,6 +648,41 @@ This endpoint manages binlogs
...
 ```
 
+## PATCH /backup/{stream_id}
+
+This endpoint updates settings of a single, already completed backup.
+
+```
+{
+  "broken": "{bool}",
+  "wait_for_applied_broken": "{number}",
```
VitaliVinahradski: doc need to be updated to match current implementation

myhoard/controller.py
```diff
@@ -1345,16 +1364,78 @@ def _handle_mode_restore(self):
...
+                if expires_at:
+                    self.state_manager.update_state(
+                        broken_backup_results=self.state["broken_backup_results"]
+                        | {request_id: BrokenBackupResult(success=succes, error=error, expires_at=expires_at)},
+                    )
+
+    def _check_existing_streams(self, stream_id: str) -> Backup | None:
+        broken_backup = self.get_backup_by_stream_id(stream_id=stream_id)
         if not broken_backup:
             raise Exception(
```
VitaliVinahradski, line 1411: nit: user gets 500 error here. Suggest using BadRequest or ValueError and fix test

myhoard/controller.py
```diff
@@ -1345,16 +1364,78 @@ def _handle_mode_restore(self):
...
+                broken_backup_request = self.state["broken_backup_requests"][0]
+            stream_id = broken_backup_request["stream_id"]
+            request_id = broken_backup_request["request_id"]
+            expires_at = broken_backup_request["expires_at"]
+            broken = broken_backup_request["broken"]
+            broken_backup = self.get_backup_by_stream_id(stream_id=stream_id)
+            if not broken_backup:
+                error = f"Stream {stream_id} does not exist anymore"
+                logging.info("Skipping marking stream=%s to broken=%s as stream does not exist anymore", stream_id, broken)
+                succes = False
```
VitaliVinahradski: nit: typo

myhoard/controller.py
danielblasina-tech: it is not correct to have BadRequest here, sine not always is a http request, did choose to catch the exception and trow again as BadRequest in request_mark_stream_as_broken

VitaliVinahradski (dismissed):
LGTM

VitaliVinahradski (approved):

PR #244, opened 2026-07-29
Keep the event scheduler off while restoring a backup

vladislav-manafov-aiven:
A restore that has binary logs to apply can get stuck on a duplicate key. myhoard restarts mysqld with the binary log disabled for the restore phases but leaves event_scheduler at its default ON, and a physical basebackup keeps events in ENABLED state. Every restored event whose schedule lapsed during the restore therefore fires within seconds of startup, and the rows it writes collide with the transactions still being replayed from the binary logs. The applier stops with HA_ERR_FOUND_DUPP_KEY and retrying does not help, because the row is already there. Rebuilding the node only runs the same lottery again.

Pass --event-scheduler=OFF on the restarts that already disable the binary log. Only the restore phases run that way, so recognising them needs no new parameter, and this has to be a startup option rather than a SET GLOBAL: by the time we could connect and issue one, the events have already fired.

The restart that finalizes a restore asks for the binary log back and so produces no options at all, which drops MYSQLD_OPTS from the systemd environment file entirely. The restored server comes back with whatever its configuration file says, rather than a hardcoded ON that would override an operator who turned the scheduler off deliberately.

The two option lists that had drifted apart are now one helper, so a change like this cannot land in the subprocess path alone while production, which goes through myhoard_mysql_env_update and systemd, keeps the old behaviour. That path had no tests at all; it has some now.

### DIFF
--- a/myhoard/myhoard.py
+++ b/myhoard/myhoard.py
@@ -7,6 +7,7 @@ from myhoard.util import (
     detect_running_process_id,
     find_extra_xtrabackup_executables,
     parse_dow_schedule,
+    restore_mysqld_options,
     wait_for_port,
 )
 from myhoard.web_server import WebServer
@@ -154,14 +155,7 @@ class MyHoard:
         if systemd_service:
             self._restart_systemd(with_binlog=with_binlog, with_gtids=with_gtids, service=systemd_service)
         else:
-            mysqld_options = []
-            if not with_binlog:
-                mysqld_options.append("--disable-log-bin")
-                # If config says slave-preserve-commit-order=ON MySQL would refuse to start if binlog is
-                # disabled. To prevent that from happening ensure preserve commit order is disabled
-                mysqld_options.append("--skip-slave-preserve-commit-order")
-            if not with_gtids:
-                mysqld_options.append("--gtid-mode=OFF")
+            mysqld_options = restore_mysqld_options(with_binlog=with_binlog, with_gtids=with_gtids)
             self._restart_process(mysqld_options=mysqld_options)
 
         # Ensure the server is accepting connections
--- a/myhoard/util.py
+++ b/myhoard/util.py
@@ -515,6 +515,25 @@ def detect_running_process_id(command: str) -> Tuple[Optional[int], bytes]:
     return ids[0], output_bytes
 
 
+def restore_mysqld_options(*, with_binlog: bool, with_gtids: bool) -> list[str]:
+    """Extra mysqld command line options for a server start that is part of a restore"""
+    options = []
+    if not with_binlog:
+        options.append("--disable-log-bin")
+        # If config says slave-preserve-commit-order=ON MySQL would refuse to start if binlog is
+        # disabled. To prevent that from happening ensure preserve commit order is disabled
+        options.append("--skip-slave-preserve-commit-order")
+        # Only restore phases run with the binlog disabled. A physical basebackup keeps events in
+        # ENABLED state and every restored event whose schedule lapsed during the restore fires
+        # within seconds of startup, so their writes race the transactions being replayed from the
+        # binlogs and the applier dies on a duplicate key. This has to be a startup option: by the
+        # time we could connect and SET GLOBAL, the events have already fired.
+        options.append("--event-scheduler=OFF")
+    if not with_gtids:
+        options.append("--gtid-mode=OFF")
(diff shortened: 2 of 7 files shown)


myhoard/util.py
```diff
@@ -515,6 +515,25 @@ def detect_running_process_id(command: str) -> Tuple[Optional[int], bytes]:
     return ids[0], output_bytes
 
 
+def restore_mysqld_options(*, with_binlog: bool, with_gtids: bool) -> list[str]:
```
danielblasina-tech, line 518: would be possible to have as parameter to disable the event_scheduler to have better control? `with_scheduler`

myhoard/myhoard.py
```diff
@@ -154,14 +155,7 @@ def _restart_mysqld(self, *, with_binlog, with_gtids):
...
         else:
-            mysqld_options = []
-            if not with_binlog:
-                mysqld_options.append("--disable-log-bin")
-                # If config says slave-preserve-commit-order=ON MySQL would refuse to start if binlog is
-                # disabled. To prevent that from happening ensure preserve commit order is disabled
-                mysqld_options.append("--skip-slave-preserve-commit-order")
-            if not with_gtids:
-                mysqld_options.append("--gtid-mode=OFF")
+            mysqld_options = restore_mysqld_options(with_binlog=with_binlog, with_gtids=with_gtids)
```
danielblasina-tech, line 158: when the service runs as systemd the scheduler is still enabled

danielblasina-tech (changes requested):

myhoard/myhoard.py
vladislav-manafov-aiven: This is still a draft. I am testing changes on my dev env right now and will come back later

danielblasina-tech: sorry didn't notice that is draft

vladislav-manafov-aiven: To the point. `_restart_systemd` does not build the options. It is done through `class EnvironmentUpdater`. It uses the same `restore_mysqld_options()`

myhoard/util.py
vladislav-manafov-aiven: Fair point in general — more control is usually a good thing. But I don't think it helps for this particular flag, and here's why.

The other two flags really do change between steps. The binary log is off while we rebuild tables and apply binlogs, and on again at the end. GTIDs are off while we patch `mysql.gtid_executed`, and on for the apply. So they have to be parameters.

The event scheduler doesn't work that way. It would be "off" for every restore step and "leave it alone" at the end, always, with nothing to decide.

The reason it can never differ: a restore only runs into an empty data directory (we refuse otherwise). So every event on the server got there from the backup. Letting them run during the restore means the backup's own events write rows into half-restored data, with the binary log off so they leave no trace, at the same moment the applier is replaying those same rows. That's exactly the failure this PR fixes. I can't think of a setup where someone would want that.

If you'd still like it spelled out, I'm happy to add a parameter to `restore_mysqld_options` itself and leave the callback and saved state alone (check the code below). That gets the explicitness cheaply. What I'd rather avoid is threading it all the way through until something actually needs to set it.

```
  def restore_mysqld_options(
      *, with_binlog: bool, with_gtids: bool, disable_event_scheduler: bool = True
  ) -> list[str]:
      """Extra mysqld command line options for a server start that is part of a restore.

      disable_event_scheduler=False leaves the event scheduler to my.cnf during the restore too.
      That is the pre-fix behaviour and lets restored events write while binlogs are being applied.
      """
      options = []
      if not with_binlog:
          options.append("--disable-log-bin")
          # If config says slave-preserve-commit-order=ON MySQL would refuse to start if binlog is
          # disabled. To prevent that from happening ensure preserve commit order is disabled
          options.append("--skip-slave-preserve-commit-order")
      if not with_gtids:
          options.append("--gtid-mode=OFF")
      # Only the restore phases run with the binary log disabled, and that is exactly when restored
      # events must not fire: a basebackup keeps them ENABLED, their lapsed schedules fire within
      # seconds of startup, and their writes race the transactions being replayed from the binlogs.
      # The finalizing restart passes no options at all, so my.cnf decides again.
      if disable_event_scheduler and not with_binlog:
          options.append("--event-scheduler=OFF")
      return options
```

danielblasina-tech (approved):

PR #245, opened 2026-07-29
fix flaky unittests

VitaliVinahradski:
- **Wait for standby to apply relay log before promoting in test**
- **Keep preserved-backup timestamps valid for whole test session**
- **Give restore more time to complete in local flow test**


vladislav-manafov-aiven (approved):
LGTM

PR #247, opened 2026-07-30
Restoration switch backup when broken

danielblasina-tech:
Broken backups were not skipped during restoration. It now checks if the backup is marked as broken and tries to switch to the previous working one.


VitaliVinahradski (approved):
LGTM

PR #248, opened 2026-08-04
Do not stop stream for broken backups

danielblasina-tech:
Previously, marking an active stream as broken also closed it. This
stopped binlog uploads and left the system unable to take new
backups. Attempting to create a new backup would then enter a loop,
repeatedly logging "marked as closed multiple times".

With this change, the stream is no longer closed when marked as
broken, avoiding this failure loop and preventing data loss.


VitaliVinahradski (approved):
LGTM

PR #249, opened 2026-08-19
Added use_memory parameter to restore options

danielblasina-tech:
Tested with sysbench with 8 tables and 20 milions rows

| use_memory        | Duration      |
|-------------|-------------------|
| 100 MB | **5m 09s** |
| 500 MB | **4m 10s** |
| 1 GB      |  **2m 15s** |
| 2 GB      | **3m 39s** |
| 4 GB      | **2m 25s** |

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,14 @@ Maximum percentage of system memory to allow xtrabackup to use while
 preparing a basebackup for restoration. If not defined, use xtrabackup's
 default value.
 
+**restore_use_memory**
+
+This option affects how much memory is allocated and is similar to
+innodb_buffer_pool_size. This option is only relevant in the --prepare phase.
+The default value is 100MB. The recommended value is between 1GB to 2GB.
+THe size is specified in bytes.
+When use_memory and free_memory_percentage are both set, free_memory_percentage will be ignored.
+
 **restore_max_binlog_bytes**
 
 Maximum amount of disk space to use for binary logs including pre-fetched logs
--- a/myhoard/basebackup_restore_operation.py
+++ b/myhoard/basebackup_restore_operation.py
@@ -52,6 +52,7 @@ class BasebackupRestoreOperation:
         target_dir: str,
         temp_dir: str,
         backup_tool_version: str | None = None,
+        use_memory: Optional[int] = None,
         prepare_progress_callback: Optional[Callable[..., None]] = None,
     ):
         self.current_file = None
@@ -70,6 +71,7 @@ class BasebackupRestoreOperation:
         self.stream_handler = stream_handler
         self.target_dir = target_dir
         self.temp_dir = temp_dir
+        self.use_memory = use_memory
         self.backup_xtrabackup_info: Dict[str, str] | None = None
         self.backup_tool_version = backup_tool_version
         # LSN bounds for prepare progress. Use last_lsn (redo stop LSN) over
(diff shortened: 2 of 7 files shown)


myhoard/basebackup_restore_operation.py
```diff
@@ -170,6 +172,8 @@ def prepare_backup(
                 # --use-free-memory-pct introduced in 8.0.30, but it doesn't work in 8.0.30 and leads to PBX crash
                 if self.free_memory_percentage is not None and get_xtrabackup_version() >= (8, 0, 32):
                     command_line.insert(2, f"--use-free-memory-pct={self.free_memory_percentage}")
+                if self.use_memory is not None:
+                    command_line.extend(["--use-memory", str(self.use_memory)])
```
alexole, line 184: Could you please check if --use-memory silently disables --use-free-memory-pct? In this case we should add config validation or make an explicit log of what is used.

danielblasina-tech: I think that documentation doesn't mention of any conflict, will do some testing/look ad the code

alexole: I think it does, if I read the code right https://github.com/percona/percona-xtrabackup/blob/52046d7ed510c97a4e0ee41efef8fa42d0be8858/storage/innobase/xtrabackup/src/xtrabackup.cc#L2346

danielblasina-tech: will check

danielblasina-tech: Now there is the check that both are present and also the log if someone call directly the class

alexole: prepare_progress_pct is the progress of prepare stage, I meant --use-free-memory-pct

danielblasina-tech: did mess with copy paste :sweat_smile: fixed

myhoard/myhoard.py
```diff
@@ -119,7 +119,8 @@ def _load_configuration(self):
             raise Exception("Either 'start_command' or 'systemd_service' must be specified")
         if start_command and not isinstance(start_command, list):
             raise Exception("'start_command' must be a list")
-
+        if self.config.get("restore_use_memory") and self.config.get("restore_free_memory_percentage"):
+            raise ValueError("Both restore_use_memory and restore_free_memory_percentage are set, but only one is allowed")
```
alexole: Because the config check is fatal, the code in restoration which checks for combination of the settings is unreachable. If _load_configuration succeeded, use_memory and free_memory_percentage can never both be set on the object, since both values come from that same config.
I would prefer not to fail here completely, rather log a warning, and build a command based on precedence in restoration and log which one was used.
Also this way we can safely reload configuration with both settings and only one will be used without hard failure.

README.md
```diff
@@ -395,6 +395,14 @@ Maximum percentage of system memory to allow xtrabackup to use while
 preparing a basebackup for restoration. If not defined, use xtrabackup's
 default value.
 
+**restore_use_memory**
+
+This option affects how much memory is allocated and is similar to
+innodb_buffer_pool_size. This option is only relevant in the --prepare phase.
+The default value is 100MB. The recommended value is between 1GB to 2GB.
+Multiple values are supported if you provide the unit (1MB, 1M, 1GB, 1G).
```
alexole: It seems it's int in the code, but here is suggests string? "1MB"

danielblasina-tech: fixed

myhoard/myhoard.py
danielblasina-tech: fixed

alexole (approved):

PR #250, opened 2026-08-19
Add lock_ddl xtrabackup setting for reduced DDL locking [MYC-179]

alexole:

Pass `--lock-ddl=REDUCED` to `xtrabackup --backup` when the new `lock_ddl`
xtrabackup setting asks for it. With the default ON the DDL lock is taken
before the data files are copied and held until the copy ends, so it lasts
as long as the backup and grows with the dataset. REDUCED copies InnoDB
data unlocked, tracks DDL through the redo log and takes a short lock at
the end only to reconcile the tables that DDL touched, which keeps the lock
window roughly flat as data grows. Measured on 27GB, the lock dropped from
~120s with ON to ~2.3s with REDUCED.

The setting defaults to ON in DEFAULT_XTRABACKUP_SETTINGS, and ON leaves
the option off the command line entirely, so the command line is unchanged
unless callers opt in.

REDUCED only exists in Percona XtraBackup 8.4 and newer, so it is gated on
the version of the binary that is about to run, which also covers the
MySQL 8.4+ requirement of the feature. An older xtrabackup, or a value we
don't recognise, logs a warning and falls back to ON instead of failing
every backup attempt. OFF is deliberately not accepted because it leaves
DDL executed during the copy unprotected.

REDUCED works for both full and incremental backups but cannot be combined
with xtrabackup's page tracking, which MyHoard does not use.

### DIFF
--- a/myhoard/basebackup_operation.py
+++ b/myhoard/basebackup_operation.py
@@ -1,7 +1,16 @@
 # Copyright (c) 2019 Aiven, Helsinki, Finland. https://aiven.io/
 from contextlib import suppress
 from myhoard.errors import BlockMismatchError, XtraBackupError
-from myhoard.util import CHECKPOINT_FILENAME, get_mysql_version, mysql_cursor, parse_xtrabackup_info
+from myhoard.util import (
+    BinVersion,
+    CHECKPOINT_FILENAME,
+    get_mysql_version,
+    get_xtrabackup_version,
+    LOCK_DDL_ON,
+    LOCK_DDL_REDUCED,
+    mysql_cursor,
+    parse_xtrabackup_info,
+)
 from packaging.version import Version
 from rohmu.util import increase_pipe_capacity, set_stream_nonblocking
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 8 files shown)


myhoard/basebackup_operation.py
```diff
@@ -182,8 +199,50 @@ def _build_backup_command_line(self, *, mysql_config_file_name: str, encryption_
...
+
+        REDUCED works for both full and incremental backups. It is incompatible with xtrabackup's page
+        tracking (`--page-tracking`), which MyHoard does not use - anything adding page tracking must not
+        combine it with REDUCED. See
+        https://docs.percona.com/percona-xtrabackup/8.4/reduction-in-locks.html
+        """
+        if self.lock_ddl in (None, LOCK_DDL_ON):
+            return self.lock_ddl
+
+        if self.lock_ddl != LOCK_DDL_REDUCED:
```
VitaliVinahradski: Small thing: this compare is case-sensitive, so `"lock_ddl": "reduced"` (or `"on"`) ends up in the unsupported branch. We then drop `--lock-ddl`, xtrabackup falls back to ON, and the operator gets the long lock they were trying to avoid. Only hint is a warning on every backup. Maybe `.upper()` the value before comparing?

alexole: You are absolutely right! I'll address it.

VitaliVinahradski (approved):
LGTM

PR #251, opened 2026-08-21
Disable scheduled backups

sc-david-voisin:
Enable (default) or disable automatic scheduled backups.

This is useful when backup timing should be fully controlled
by an external scheduler rather than MyHoard itself. Note that this only
affects new base backup scheduling; binary log uploads and existing backup
streams continue as usual.

### Unit tests

```
$ hatch run python -m pytest -q test/test_controller.py                                        

.....s............................................................................                                                                                                     [100%]
====================================================================================== warnings summary ======================================================================================
test/test_controller.py: 520 warnings
  /usr/lib/python3.12/contextlib.py:137: DeprecationWarning: 'db' is deprecated, use 'database'
    return next(self.gen)
...

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
81 passed, 1 skipped, 522 warnings in 503.20s (0:08:23)
```

### DIFF
--- a/test/test_controller.py
+++ b/test/test_controller.py
@@ -1842,6 +1842,46 @@ def test_periodic_backups_are_paused(time_machine, master_controller) -> None:
     while_asserts(lambda: streaming_binlogs(m_controller, 2), timeout=10)
 
 
+def test_scheduled_backups_can_be_disabled(time_machine, master_controller) -> None:
+    # pylint: disable=protected-access
+    time_machine.move_to("2023-01-02T18:00:00")
+
+    # By default backup_hour = 3, backup_interval_minutes = 1440
+    m_controller, master = master_controller
+
+    m_controller.switch_to_active_mode()
+    m_controller.start()
+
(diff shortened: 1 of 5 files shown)


test/test_controller.py
```diff
@@ -1842,6 +1842,46 @@ def flush_binlogs():
     while_asserts(lambda: streaming_binlogs(m_controller, 2), timeout=10)
 
 
+def test_scheduled_backups_can_be_disabled(time_machine, master_controller) -> None:
+    # pylint: disable=protected-access
```
alexole: This is probably not needed. Could you remove it?

alexole (changes requested):
Thanks for the contribution.
Could you fix the failing tests? And also left a small comment, otherwise LGTM.

test/test_controller.py
sc-david-voisin: done in 6487a6d135fd8bd192f3216eda4d8681c581639d

alexole (approved):

PR #252, opened 2026-09-11
Fold out of order GNOs into one GTID range when scanning binlogs

VitaliVinahradski:
## Why

myhoard summarises every binlog into `gtid_ranges`, one range per stretch of consecutive GNOs, and puts that list into the upload metadata of the file. Two things MySQL does break the "consecutive" assumption:

- The parallel applier of a replica can write adjacent transactions swapped into the replica's own binlog. This bites after a node replacement or failover, when the new primary has to upload a binlog it wrote as a replica.
- A node that replaces the primary replays the old primary's binlogs and, before promotion, writes transactions of its own into the same file. The two servers take turns, 87 times in the case from 2026-09-20.

Either way the list grows to hundreds of tiny ranges, the metadata outgrows the 2 KB object storage limit and the upload of that binlog fails forever. Backups go stale and binlogs pile up on disk until an operator rewrites the stream state (MYC-211).

## What

`build_gtid_ranges` now folds the GNOs of each server as if they were sorted, over the whole file: one range per server and per real gap. Both cases above come out as one or two ranges. Files where every server appears once, the normal case, produce exactly the list they did before.

The fold stays streaming. In-order events extend the current range at one comparison each, a GNO that does not touch it waits in a pending range keyed by its start and end GNO, and neighbours merge in constant time. Nothing to tune. On 340k events, a 64 MiB binlog of minimal transactions:

| input | before | this PR |
|---|---|---|
| in order | 0.04 s | 0.08 s |
| one swap every 500 | 2038 ranges | 1 range |
| every block of 8 shuffled | 301k ranges, 79 MiB | 1 range |
| 340k real gaps | 340k ranges, 89 MiB | 340k ranges, 124 MiB |

Ranges now come in GNO order per server instead of file order, and `start_ts`/`end_ts` are the earliest and latest timestamp in the range. Two spots in the restore coordinator relied on file order and are fixed: the PITR skip check and listing look at all ranges of a file, not the first one, and the apply completion check waits for every range of the last binlog of a batch, not only the last range. State files written by an older myhoard are still read.

## Tests

Edge cases (both ends of a file, GNO 1 and the largest GNO, timestamps going backwards inside a file, a swap across two files, two servers alternating, an older state file), hypothesis properties against a sorted model, the 87 runs of the 2026-09-20 binlog as a regression test, and a minimal binlog writer so `read_gtids_from_log` is testable.

One known limitation is pinned as a strict xfail: the reader stops at the first event at or after the PITR target time, so a transaction that committed before the target but was written after one at the target is left out. Reading on would apply transactions past the target instead, and a hole before the target is the better of the two.

Streams already stuck on an oversized range list keep their stored ranges and still need the manual fix from the runbook.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


myhoard/util.py
```diff
@@ -220,35 +221,109 @@ def read_gtids_from_log(
...
+            left["end"] = gno
+            left["end_ts"] = timestamp
+            if right and right["start"] == gno + 1:
+                # The last hole between the two ranges is filled, join them
+                left["end"] = right["end"]
+                left["end_ts"] = right["end_ts"]
+                del ranges[slot]
+        elif right and right["start"] - 1 == gno:
+            right["start"] = gno
+            right["start_ts"] = timestamp
```
alexole: This changes what start_ts means and it breaks a PITR check.

Before, ranges were built in file order and commit timestamps grow in file order, so start_ts was the earliest timestamp in the range and end_ts the latest. Now they are the timestamps of the lowest and highest GNO events, so the pair is no longer an envelope over the range.

restore_coordinator.py:1197 relies on it being an envelope. Its own comment says a file whose start_ts is at or after target time is "certain not to contain data we're going to apply", and when that trips it also calls target_time_reached_by_server.add(), which skips every later binlog from that server too.

Take events (ts 1000, gno 2), (ts 1001, gno 1), (ts 1002, gno 3). You get start=1, end=3, start_ts=1001. The transaction at ts 1000 is inside the range but outside [start_ts, end_ts]. A restore with target_time 1001 now skips this binlog and all following ones from that server, and that transaction is silently lost. The lowest GNO can arrive up to GTID_REORDER_WINDOW positions late, so on a slow writer this gap can span seconds. Line 1209 has the mirror problem: end_ts can understate the file's real last timestamp and leave target_time_reached unset for the file that actually holds the target.

Could we keep start_ts as the min and end_ts as the max over the events folded into the range? _model_gtid_ranges and the "swap at file start" case bake in the current semantics, so they would need updating with it.

myhoard/restore_coordinator.py
```diff
@@ -810,7 +810,9 @@ def apply_binlogs(self) -> None:
             ranges = list(build_gtid_ranges(read_gtids_from_log(file_name, read_until_time=self.target_time)))
             if ranges:
                 last_range = ranges[-1]
-                until_after_gtids = f'{last_range["server_uuid"]}:{last_range["end"]}'
+                # Stop once every GTID in the file is applied. The end of the last range would not be enough
+                # because a replica can write a transaction into its binlog after a transaction with a higher GNO.
+                until_after_gtids = make_gtid_range_string(ranges)
```
alexole, line 819: The UNTIL clause now covers the whole set, but the completion check still waits for a single range, so the two disagree and the check can finish early.

last_range = ranges[-1] on line 812 is unchanged. It becomes current_executed_gtid_target on line 926, and _check_sql_replica_status builds range_str = make_gtid_range_string([expected_range]) on line 1666 from that one range. Before this change ranges[-1] was the last range in file order, which is the right thing to wait for. After folding it is the highest GNO range instead.

Line 820 sets relay_log_target = None on this path, so expected_index is None and the relay index guard on line 1627 does not apply. The GTID_SUBSET check is all that is left.

Take a file with GNO 1..100, then 105, then 101, which is exactly the reordering this PR is about. Folding gives [{start:1,end:101}, {start:105,end:105}]. until_after_gtids is correct at 1-101:105. But current_executed_gtid_target is {105,105}, and GTID_SUBSET('a:105', @@GLOBAL.gtid_executed) turns true as soon as 105 is applied, which is before 101 in file order. found fires, STOP REPLICA on line 1690 cuts the apply short, and line 967 then adds the binlog's full gtid_ranges to gtid_executed. GTID 101 is never applied but is recorded as executed.

Either make current_executed_gtid_target hold all the ranges and check the whole set, or keep it pointing at the file order last range so its old meaning survives. The state field is a single Optional[GtidRangeDict] today, so the first option needs a small type change.

Worth noting this path has no test coverage, nothing in test/ mentions SQL_AFTER_GTIDS or current_executed_gtid_target.

myhoard/util.py
```diff
@@ -220,35 +221,109 @@ def read_gtids_from_log(
...
+        else:
+            ranges.insert(slot, _new_gtid_range(ranges[0]["server_id"], ranges[0]["server_uuid"], gno, timestamp))
+
+    def close_excess(self) -> List[GtidRangeDict]:
+        """Close all but the highest range once more than GTID_REORDER_WINDOW are open, so memory stays bounded"""
```
alexole: Two notes on the cap, both minor compared to the other comments.

Once close_excess fires, flushed ranges can no longer be extended, so a later GNO that would have joined one emits a separate adjacent range instead. That re-introduces the fragmentation this PR removes, and the output stops being in GNO order, which the docstring on line 235 promises. With the window shrunk to 3, input [1,3,5,7,4] comes out with starts 1,3,5,4,7.

Second, this bounds the ranges held in memory but not the count that ends up stored, since flushed ranges are yielded and kept. A binlog with a few hundred real gaps still produces a gtid_ranges list that can outgrow the object storage metadata limit, which is the original failure from a different source. Filtered transactions would do it, or many server uuid runs after a chain of node replacements, which your many_servers_in_sequence test already models at 40. Probably a separate PR, but a guard on the stored count somewhere would close the class off.

On the window itself: the distance test in fits() does not actually bound anything, since 1,3,5,7,... passes it every time and only close_excess stops the growth. And when the distance test does fire it can make the output bigger. [1, 2, 1030, 3] gives (1-2), (1030-1030), (3-3), where leaving them open gives (1-3), (1030-1030). Would it be simpler to drop the distance check and keep just the cap on open ranges, renamed to something like MAX_OPEN_GTID_RANGES?

alexole (commented):
Had a close look at this because we hit the same thing on a MySQL 8.4.8 HA pair. The folding itself looks right to me. I fuzzed it against a sorted model and coverage is exact, nothing is dropped, and start_ts/end_ts always belong to a real event with that GNO.

Two things I think need fixing first, both coming from the same place: the change alters what the gtid_ranges fields mean, and one consumer was updated but two others were not. The start_ts/end_ts change breaks the point in time recovery skip check in restore_coordinator, and ranges[-1] no longer means what current_executed_gtid_target expects, which can stop a restore early and record GTIDs as executed that were never applied. Details inline.

myhoard/util.py
VitaliVinahradski: Agreed. Covered by tests and fixed in the new revision: start_ts and end_ts are min and max over the events of the range, and the listing check now looks at all ranges of a file, not only the first.

myhoard/restore_coordinator.py
VitaliVinahradski: Confirmed with your exact input, 1..100 then 105 then 101. The test drives apply_binlogs and _check_sql_replica_status with a fake cursor whose GTID_SUBSET has 105 applied and 101 not, and on the old code the check reported done. current_executed_gtid_target now holds all ranges of the last binlog of the batch, same set the UNTIL clause is built from on the PITR path and the full gtid_ranges of the last binlog otherwise, and GTID_SUBSET runs over the whole thing. A state file from an older myhoard has a single range there and gets wrapped in a list, there is a test for that too.

myhoard/util.py
VitaliVinahradski: Both gone. You were right that the cap bounded the open ranges but not the stored ones, and both callers hold the full list anyway, so it bought nothing and cost GNO order when it fired. The builder now keeps the range in-order events extend as the current one and parks stray GNOs in pending ranges keyed by start and end GNO, so neighbours merge in O(1), nothing is ever flushed early and the output is the sorted GNOs of the run for any order. Your [1,3,5,7,4] and [1,2,1030,3] inputs are tests now. On the stored count guard I agree it is a separate PR: filtered transactions or a long chain of server uuids in one file still make real gaps that folding cannot remove, and what to do then, collapse, mark the stream broken or move the ranges out of the header, is its own discussion.

VitaliVinahradski:
Verified end to end on a dev environment with this PR's head (ba541b3) built into the myhoard package.

Node replacement under load (4 applier workers, load of concurrent no-op `CREATE TABLE IF NOT EXISTS` plus inserts, the trigger for MYC-211): the replaced node's catch-up binlog held 15230 GTIDs from the old primary with 1481 out of order pairs in file order. The scanner produced one range for them, plus one for the new server's uuid, and the upload went through first try. The old builder would have split that file into thousands of ranges.

Point in time fork with a target inside that binlog: the restore computed the stop set `…:1828-16674`, applied it, and the fork's `gtid_executed` matched it exactly. That stop set came out as a single range, so the multi-range `UNTIL SQL_AFTER_GTIDS` path from the second commit was not hit by this run. The unit tests cover it.

The same test surfaced a pre-existing PITR problem that is unrelated to this PR. Tracked separately as MYC-304.

<details><summary>Evidence</summary>

Scanner on the replaced node, the catch-up binlog (4778363 bytes):

```
found 2 GTID ranges: [{'end': 17057, 'end_ts': 1789566194, 'server_id': 3404109513, 'server_uuid': '094ecd80-…', 'start': 1828, 'start_ts': 1789565872}, {'end': 15, 'end_ts': 1789566198, 'server_id': 249044342, 'server_uuid': '80761472-…', 'start': 1, 'start_ts': 1789566197}]
Successfully uploaded local binlog index 2 as remote index 4 in 0.2 seconds
```

File order of the old primary's GTIDs in that file, from mysqlbinlog:

```
gtids=15230 descents=1481 forward_jumps=2600 max_gno=17057
descent at gno 1834 after 1835
descent at gno 1839 after 1840
descent at gno 1846 after 1847
```

PITR fork, target 1789566185:

```
Restoring up to and including target GTID '094ecd80-…:1828-16674'
Expected log file 'relay.000004' reached and GTID set '094ecd80-…:1828-16674' has been applied: 094ecd80-…:1-16674
Backup restoration completed
```

Fork `@@gtid_executed`: `094ecd80-…:1-16674` plus the fork's own uuid.

</details>

myhoard/restore_coordinator.py
```diff
@@ -1192,27 +1198,31 @@ def _list_binlogs_in_bucket(
                     if binlog["server_id"] in target_time_reached_by_server:
                         continue
                     if self.target_time and binlog["gtid_ranges"]:
-                        if binlog["gtid_ranges"][0]["start_ts"] >= self.target_time:
+                        # A file holds one range per server and per real gap, in GNO order per server, so the
+                        # earliest and the latest commit time of the file can sit in any of them
+                        first_ts = min(rng["start_ts"] for rng in binlog["gtid_ranges"])
+                        last_ts = max(rng["end_ts"] for rng in binlog["gtid_ranges"])
```
alexole: **Correctness (high): the file-wide `max(end_ts)` makes the PITR early-stop fire on a replayed-history binlog and silently truncate the restore.**

Now that ranges are folded per server over the whole file, the timestamps of one binlog can come from two different clocks: the transactions a node replays from its predecessor keep the *source's* commit times, while the transactions the node writes itself carry the current time. That is exactly the file this PR is built around (`RUNS_OF_A_REPLACED_PRIMARY`; `test_build_gtid_ranges_two_servers_alternating_in_one_file` shows both folded ranges spanning almost the whole file).

Scenario: node replaced at 12:00. While it replays the old primary's 09:00-12:00 history into its own binlogs #1..#5, it also writes its own transactions at 12:00-12:10. Binlog #1 therefore holds replayed events at 09:00-09:30 plus own events at 12:00-12:05. With `target_time = 11:00`:

* `last_ts = max(end_ts) = 12:05 >= 11:00`, so `server_id` goes into `target_time_reached_by_server` (line 1217, the sibling site of this hunk).
* Binlogs #2..#5 are then dropped by the `if binlog["server_id"] in target_time_reached_by_server: continue` guard above, and `target_time_reached=True` is persisted, so `_fetch_more_binlog_infos` stops polling.
* The final round computes its `UNTIL SQL_AFTER_GTIDS` from binlog #1, whose reader stops at the first event at or after 11:00, i.e. at the 12:00 event. The restore finalizes having applied history only up to 09:30 and loses the 09:30-11:00 replayed transactions, with no error anywhere.

This is a regression from the `max`, not from the folding: the old `binlog["gtid_ranges"][0]["end_ts"]` read the *replayed* run's old timestamp (~09:00) and kept fetching, and with folding alone `gtid_ranges[0]["end_ts"]` would still be ~09:30. The `first_ts = min(start_ts)` half is fine, it only makes the skip less likely.

Fix direction: don't conclude "target time reached" from a file-wide max when the file's ranges span more than one `server_uuid` (the final round's `UNTIL SQL_AFTER_GTIDS` already provides the precision), or track reached-ness per range `server_uuid` rather than per uploading `server_id`.

test/test_out_of_order_gtid_ranges.py
```diff
@@ -0,0 +1,608 @@
...
+    ],
+)
+def test_ranges_are_as_compact_as_the_sorted_gnos_and_in_gno_order(gnos, open_range_cap):
+    """build_gtid_ranges promises ranges in GNO order and folds GNOs written out of order into one range.
+    The output must be the sorted GNOs of the run collapsed into ranges, no matter how the builder bounds
+    its memory. The cap is lowered here so that a five event input hits it."""
+    if open_range_cap is None:
+        ranges = _ranges(gnos)
+    else:
+        with patch.object(myhoard_util, "GTID_REORDER_WINDOW", open_range_cap, create=True):
```
alexole: **Test coverage (low): this `open_range_cap` case is vacuous.**

`myhoard_util.GTID_REORDER_WINDOW` does not exist in `myhoard/util.py` (hence the `create=True`), and nothing in the builder reads such a name, so the patch has no effect. The `"closing ranges at the cap must not split or reorder them"` case runs exactly the same code path as the unpatched one, and the docstring line "The cap is lowered here so that a five event input hits it" is not true. Either drop the parametrization or point it at a knob the builder actually consults.

myhoard/restore_coordinator.py
```diff
@@ -1610,7 +1627,10 @@ def _start_sql_thread_if_not_running(self, replica_status: ReplicaStatus) -> Non
             restart_unexpected_dead_sql_thread(cursor, replica_status, self.stats, self.log)
 
     def _check_sql_replica_status(self) -> tuple[bool, int]:
-        expected_range = self.state["current_executed_gtid_target"]
+        expected_ranges = self.state["current_executed_gtid_target"]
+        if isinstance(expected_ranges, dict):
```
alexole, line 1631: This seems to be redundant as node/server is allowed to restore once (transition from idle state to restore), state file is local, so never taken the old one.

VitaliVinahradski: this is open source code and we don't know how it is used outside. I suggest keeping this code

VitaliVinahradski: deleted this code

